PR #1, opened 2019-02-20
adding lesson01 assignment

njschafi:
Adding lesson01 files and test files.

### DIFF
--- a/students/njschafi/Lesson01/inventory_management/electric_appliances_class.py
+++ b/students/njschafi/Lesson01/inventory_management/electric_appliances_class.py
@@ -1,0 +1,28 @@
+"""Extension of inventory class. Creates an electric appliance object
+with inputted attributes from user."""
+#pylint: disable-msg=too-many-arguments
+# pylint: disable=too-few-public-methods
+from inventory_class import Inventory
+
+
+class ElectricAppliances(Inventory):
+    """Creates instance of an electric appliance. User inputs product_code,
+    description, market_price, rental_price, brand, voltage."""
+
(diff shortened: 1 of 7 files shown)


ldconejo:
Hi Neima,

For this one, where you able to run your unit tests and get them to pass as they are? I was able to after a few modifications, but not as they were.

Your integration test has a few bugs that would prevent it to work, so I assume you didn't run that one. 

Finally, once your unit test file is modified to work, coverage _main.py_ is at 81%, which doesn't meet the requirement of 90% coverage for each individual file.

Let me know if you have any questions, we can setup an Office Hours session to go over you code.

Thanks,

Luis

njschafi:
Hi Luis,

Thanks for the feedback. I was struggling for awhile on how to explicitly import modules (my unittest would fail if I ran it in another folder) so I decided to put all the files into one folder and run my tests that way. My mistake was to alter the code to explicit imports before pushing to the repo and that is why my code did not work. 

As for my unittest for main only achieving 81%. I spent hours trying to figure out why some lines weren't being tested (even though I had written tests for them). After spending some time today I realized that my unittest stopped once it hit "test_exit_program" - so it did not test any of the code thereafter. 

I have refactored my code and pushed it to the repo. It should be all be working now and main has test coverage of 94%. If possible I was wondering if there was any way I could get a regrade on this assignment? I would really appreciate it.

Thanks! 
Here is the pull request for my refactored code: 
https://github.com/UWPCE-PythonCert-ClassRepos/SP_Python220B_2019/pull/2

PR #3, opened 2019-03-01
Adding lesson02 assignment and activity files

njschafi:
Some notes regarding assignment 2:

- Ran charges_calc.py on pylint and got a 10/10.

- I did not totally understand what was needed for this task: "Your Python debugger logfile. Include it as debugger_log.txt". I did an initial debug run when I first started on the assignment and uploaded that. Please let me know if you were looking for something else.


students/njschafi/Lesson02/assignment/charges_calc.py
```diff
@@ -0,0 +1,135 @@
...
+            logging.warning('Negative amount of rental days. Product: %s',
+                            value['product_code'])
+            logging.debug('Amount of Days: %s', value['total_days'])
+
+        try:
+            value['total_price'] = value['total_days'] * value['price_per_day']
+            # can't take sqrt of negative number
+            value['sqrt_total_price'] = math.sqrt(value['total_price'])
+        except ValueError as err:
+            # logging.error('total_price is negative:'
```
ldconejo, line 103: This won't be flagged by Pylint as an issue, nevertheless, always avoid leaving commented code in a release (or a submission, in this case). The number one risk is that someone uncomments it by error later on, and it creates unnecessary clutter in what is otherwise very high quality coding.

ldconejo:
Hi Neima,

Good job on this one. The code is running and returning both an output a log messages and configurable levels. I added one comment on leaving commented code in a submission, aside from that, very good implementation.

For the Python Debugger requirement, the idea was for you to use pdb initially to figure out what issues were present in the code (by using line-by-line execution, breakpoints, etc.). I'm not going to ask you to do that at this point, since you're already well aware on what issues the original code had, so it won't add any value.

Keep up the good pace!

Luis

PR #6, opened 2019-03-12
Adding Lesson03 assignment and activity files

njschafi:
- Assignment files all have 10/10 grade from pylint.
- No unittest written for create_customer.py and customer_model.py since they are tested through basic_operations.py 
- Code passes all tests but throws errors due to raised errors. How do I prevent this from happening?


students/njschafi/Lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,115 @@
...
+        self.assertEqual(user_2['name'], query.customer_name)
+        self.assertEqual(user_2['lastname'], query.customer_last_name)
+        self.assertEqual(user_2['home_address'], query.customer_address)
+        self.assertEqual(user_2['phone_number'], query.customer_phone)
+        self.assertEqual(user_2['email_address'], query.customer_email)
+        self.assertEqual(user_2['status'], query.customer_status)
+        self.assertEqual(user_2['credit_limit'], query.customer_limit)
+
+        # add a duplicate person
+        self.assertRaises(ValueError, add_customer(**user_2))
```
ldconejo, line 66: Doing _assertRaises_ will cause your test to fail, since the exception is not contained. Instead, do it as a context manager:

        with self.assertRaises(ValueError):
            add_customer(**user_2) 

Your test will verify the correct exception is raised, but the exception itself will be "trapped" by the context manager.

students/njschafi/Lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,115 @@
...
+        query = Customer.get(Customer.customer_id == user_1['customer_id'])
+        self.assertEqual(5000.00, query.customer_limit)
+
+        # Test for non-existant customer
+        update_customer_credit('456879', 5000.00)
+        self.assertRaises(ValueError,
+                          update_customer_credit('456879', 5000.00))
+
+        # Test for non-float value inputted
+        self.assertRaises(TypeError,
```
ldconejo, line 104: Same issue here, the exception will cause the test to be reported as failing (not the case, since the exception is indeed raised). Change it to a context manager.

students/njschafi/Lesson03/assignment/test_basic_operations.py
```diff
@@ -0,0 +1,115 @@
...
+        """Tests if customer credit is updated properly"""
+        create_empty_db()
+        add_customer(**user_1)
+        update_customer_credit(user_1['customer_id'], 5000.00)
+        query = Customer.get(Customer.customer_id == user_1['customer_id'])
+        self.assertEqual(5000.00, query.customer_limit)
+
+        # Test for non-existant customer
+        update_customer_credit('456879', 5000.00)
+        self.assertRaises(ValueError,
```
ldconejo, line 100: Change to a context manager to get the test to pass.

ldconejo:
Hi Neima,

Looks very. Code is impeccable, all three files report 10/10 on Pylint.

For your testing, the way you're using _assertRaises_ causes the tests to fail, I've added a suggestion to update those tests to use a context manager. Once that is fixed, coverage is very good, 90% in one case, 100% in the other. Please fix the _assertRises_ tests to get the full grade. Submit a new pull request once that's done.

Regards,

Luis

njschafi:
Thanks for the feedback Luis! I've been wondering how to suppress those
errors during Unittest.

I refactored my code and did a new pull request
https://github.com/UWPCE-PythonCert-ClassRepos/SP_Python220B_2019/pull/7.

-Neima

PR #13, opened 2019-03-21
unittest error

vvinodh:


ldconejo:
Hi Vinodh,

One easy solution is to add the path to your _inventory_management_ folder on line 4 of _test_unit.py_. For example (you'll need to replace the path with the one in your computer)::

```
import sys
sys.path.append('/Users/ldconejo/Documents/UW/SP_Python220B_2019/SP_Python220B_2019/students/vvinodh/Lesson1/inventory_management/')
```

Let me know if that works. We can discuss other alternatives later on.

-Luis

vvinodh:
Luis,
Thank you so much. This works.

PR #14, opened 2019-03-22
Lesson02 Submission

dfspray:


students/dfspray/Lesson02/assignment/charges_calc.py
```diff
@@ -114,7 +114,8 @@ def calculate_additional_fields(data):
             value['unit_cost'] = value['total_price'] / value['units_rented']
         except ValueError as ex:
```
ldconejo, line 115: This one is open to discussion, as sometimes you will not have control over the quality of your input data and will need to live with some entries triggering exceptions. In cases in which the input data is expected not to contain these issues, it's good to log them as errors.

ldconejo:
Pylint is reporting 10/10, logging levels are correctly implemented. Good job.

PR #16, opened 2019-03-26
Lesson 2

LAFiorizarry:
When I run this using the Spyder IDE, it creates new file handlers over and over, and logs the same errors multiple times. I have to restart the console or remove each file handler. 
But when I just run in a command line it seems to work fine. Let me know if this happens when you run.


ldconejo:
Hi Laura,

Looks good. One point open to discussion would be whether issues not triggering an exception should be logged as warnings and not as errors, but that's very subjective in this case and would indeed change one way or the other in a real world application.

Pylint rated your code at 10 out 10. I look forward to your next submission.

Regards,

Luis

PR #17, opened 2019-03-28
Assignment 1 submission

vvinodh:


ldconejo:
Pylint reports 10 out 10, test coverage is higher than 90% for all files. Good job!

-Luis

PR #20, opened 2019-03-29
Lesson03 Submission

dfspray:


dfspray:
A couple of questions on the assignment for lesson04: It states that there will be "instructor provided customer data." Where can I find that? And how many comprehensions/iterators/generators should I be adding to the lesson04 assignment? I can only think of one place to use a iterator/comprehension without adding additional functionality to basic_operations and test files.

ldconejo:
Hi Daniel,

Let me check with the academic team on the instructor-provided data. I don't believe that's something required for this assignment, but I will check anyway. On the implementation of comprehensions and other structures, the requirements are open, meaning that it's up to you whether only a few or all elements can be added in a way that makes your code better. On submission, please indicate which changes you made versus your work on lesson 3.

Regards,

Luis

ldconejo:
Hi Daniel,

Code looks good, Pylint is grading it at 10. Coverage is low for _basic_operations.py_:

```
$ python -m coverage report
Name                           Stmts   Miss  Cover
--------------------------------------------------
src/__init__.py                    0      0   100%
src/basic_operations.py           95     19    80%
src/create_customer.py             7      0   100%
src/customer_model_schema.py      16      0   100%
--------------------------------------------------
TOTAL                            118     19    84%
```

Now, the assignment does not require a certain coverage, I usually recommend both trying to get at least 90% and, more important, try to make sure no key flows are left uncover. All of the missing coverage is in the _Except_ blocks of _basic_operations.py_. Try to add coverage for those blocks, at least until you get your code to 90% coverage for that file.

Regards,

Luis

PR #24, opened 2019-04-08
Lesson05 Submission

dfspray:
Could I please get my lesson04 submission graded? Also, this program assumes that I am creating my own csv files and formatting, which I put in a separate program called 'create_csv.py'. I wrote tests for that file as well.

Thanks,
Daniel


dfspray:
For Lesson06, it states that "The assignment repository for this lesson contains a file with 10 records and a module that reads that file." However, the data folder is empty, and there are no other files I could find related to this note. I ended up writing my own .csv file that could be run by the program. I also wasn't sure how far we needed to go to improve the program speed (cython/pypy/etc.), but I reduced the run time by over half using regular python. The Lesson06 files are basically done in this commit, but I wanted to wait for clarification.

ghost:
I'll check why the data file is empty. And I'm grading all the work today and updating edx

PR #29, opened 2019-04-19
Lesson 4

LAFiorizarry:
submission1


students/lauraannf/lessons/lesson04/assignment/basic_operations.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def add_customer(customer_id, first_name, last_name, home_address,
+                 phone_number, email_address, status, credit_limit):
+    """ adds customer to database"""
+    print('Adding new customer, Customer ID {}...'.format(customer_id))
+    try:
+        Customer.get_by_id(customer_id)
+        print('Customer ID {} is already in use'.format(customer_id))
+    except Exception:
```
ldconejo, line 27: Avoid using "catch all" exception handling. The problem is that it will intercept the exceptions you can think of and the ones you can't, so you code could end up catching (and incorrectly handling) an unexpected exception. Please make sure you refactor all instances of catch all exception handling with specific exceptions.

students/lauraannf/lessons/lesson04/assignment/customer_model.py
```diff
@@ -0,0 +1,37 @@
...
+"""
+Created on Mon Apr  1 15:41:41 2019
+
+@author: Laura.Fiorentino
+
+Lesson3 Assignment
+"""
+
+
+from peewee import *
```
ldconejo, line 11: This is also a good opportunity for improving your code. Try to avoid wildcard imports, in particular with large modules such as Peewee, you will end up, like in this case, with a lot of unused imports. Instead, import only those methods you will be using.

ldconejo:
db.log looks good, coverage is close to 100% for all relevant files. There are a few things that can still be refactored in your code. Take a look at that and submit an updated pull request.

LAFiorizarry:
Hi,
I'm not sure what to do when an exception doesn't have a specific name like
ValueError. For these database ones, it doesn't even say exception, just
something like CustomerDoesNotExist:. But if I use that as the exception
name it says it's undefined.What;s the best way to handle all of these?
Thanks,
Laura

ldconejo:
Hi Laura,

All exceptions will have a name, you can add this code to try and capture that name, so that you can then capture that exception in particular:

```
try:
    ***code that could trigger an exception***
except Exception, e:
    print 'Exception type:', e.__class__.__name__
```

If the code _does not_ cause an exception, such as in the case you mention, then don't use the _try/except_ block. The idea is that, should you code runs into a use case that you haven't seen before and that causes an error, you want the code to crash, not hide it, so that you are made aware of the issue and decide on the best way to fix it.

Hope this helps :-)

-Luis

LAFiorizarry:
Thanks, some of those error names were pretty hidden. Like the
CustomerDoesNotExist one is in the database. So I did except
Customer.DoesNotExist...
And the IntegrityError when there is a check contraint, that's part of
peewee. I think I got them all now. I'll re submit.

PR #38, opened 2019-04-29
Lesson07 submission

dfspray:


ldconejo:
Hi Daniel,

Excellent work on lesson 7. I see you included additional commits for lessons 8 & 9, but don't yet see a submission on the course website, so I imagine there is still some work pending on those two. Keep up the good work!

-Luis

PR #40, opened 2019-05-01
Lesson 02 submission

chelseamsmith:


ldconejo:
Hi Chelsea,

Implementation looks good. I would have added more variety to which log messages would be warnings and which would be errors (for example, missing date should, in my opinion, be logged as an error). Of course, in a real-world application, what is considered an error and a what is a warning depends on the project requirements.

Code gets a 10 out of 10 grade from Pylint, good coding practices.

Keep up the good work!

Luis

PR #41, opened 2019-05-01
Lesson08 and Lesson09 Complete

dfspray:
For Lesson09, the files I had were .png, so I modified the program to reflect that. I also added a filter to only add paths that contained .png files. The majority of the commits for Lessons08 and 09 are in pull request 38:
https://github.com/UWPCE-PythonCert-ClassRepos/SP_Python220B_2019/pull/38

### DIFF
--- a/students/dfspray/Lesson08/src/rented_items.csv
+++ b/students/dfspray/Lesson08/src/rented_items.csv
@@ -4,15 +4,3 @@ Alex Gonzales,QM15,Queen Mattress,17
 Susan Wong,LR04,Leather Sofa,25.00
 Susan Wong,KT78,Kitchen Table,10.00
 Susan Wong,BR02,Queen Mattress,17.00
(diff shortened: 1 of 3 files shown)


ldconejo:
Excellent work on assignment 8:
- All requirements met.
- Pylint graded your code at 10 out of 10.
- Coverage from your tests is at 100%.

-Luis

ldconejo:
On assignment 9, _test_create_csv.py_ has this on line 45:

`file_list = ["\\product_data.csv", "\\customer_data.csv", "\\rentals_data.csv"]`

This will work fine on a Windows system but fail on Linux or Mac OS. If you needed your application to be platform-agnostic (or just to make it more robust if your tests were later reused on another platform) you could _sys.platform_ or _os.name_ depending on whether you  need to tell Linux from MacOS from Unix or not.

No need to make any changes for this one, just something to consider when implementing code that might run on more than one OS.

Regards,

Luis

PR #42, opened 2019-05-02
Files for assignment 1

vmedina1014:


ldconejo:
Hi Victor,

Code looks good, a couple of things to improve before getting the full grade:

- Check the issues reported by Pylint. For _too-few-public-methods_, you can tell Pylint to ignore it (you can use # pylint: disable=R0903 at the beginning of your code to do this). Other issues like unused arguments should be fixed.
- Improve coverage for _main.py_, it's currently at 83%. You can run _coverage html_ for the coverage module to produce an html version of your coverage report and see what lines of code are not being covered. The goal will be to get coverage for that file past 90%.

Regards,

Luis

PR #44, opened 2019-05-03
Lehning Assignment01

KLehning:
Complete Assignment01 for Kyle Lehning

### DIFF
--- a/students/kyle_lehning/Lesson01/test_integration.py
+++ b/students/kyle_lehning/Lesson01/test_integration.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+"""
+Integration tests for electric_appliances_class, furniture_class, inventory_class,
+main, and market_prices
+"""
+
+from main import *
+from unittest import TestCase
+from unittest.mock import patch
+import io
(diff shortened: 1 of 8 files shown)


students/kyle_lehning/Lesson01/test_integration.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
+"""
+Integration tests for electric_appliances_class, furniture_class, inventory_class,
+main, and market_prices
+"""
+
+from main import *
```
ldconejo, line 7: In general, and for testing in particular, avoid wildcard imports. For all applications, it causes every single piece of code, including code that you're not planning to use, to be loaded into memory, which my impact performance when you have multiple imports and/or very large modules.

For testing, especially when testing larger projects, it's easier if a test tells you that it's failing when running _module_name.do_something()_ than just _do_something()_. The latter would make it more difficult to identify the module under test that the failing code came from.

ldconejo:
All assignment requirements met. See me comments for improvements on future submissions.

-Luis

PR #45, opened 2019-05-03
Lesson 06

LAFiorizarry:


students/lauraannf/lessons/lesson06/assignment/write_csv.py
```diff
@@ -0,0 +1,52 @@
...
+    date1 = date(2000, 1, 1)
+    new_date_random = date1 + timedelta(rand)
+    return new_date_random.strftime("%m/%d/%Y")
+
+
+def ao_choice(choice):
+    """ pick ao or no ao"""
+    if choice == 1:
+        return 'ao'
+    else:
```
ldconejo, line 35: This _else_ is unnecessary, as it is preceded by a _return_ statement, so the only way _return ''_ could happen is if _return 'ao'_ didn't. No need to fix this one, just something to keep in mind when using if statements that contain a return.

students/lauraannf/lessons/lesson06/assignment/good_perf.py
```diff
@@ -0,0 +1,59 @@
...
+        year_count = {
+            "2013": 0,
+            "2014": 0,
+            "2015": 0,
+            "2016": 0,
+            "2017": 0,
+            "2018": 0
+        }
+        for row in reader:
+            if row[5] > '00/00/2012':
```
ldconejo, line 27: A simpler way to count occurrences per year: Capture the year and see if it's present in your _year_count_ dictionary:

```
            if row[5][6:] in year_count:
                year_count[row[5][6:]] += 1
```

ldconejo: This also translates into ~30% reduction from your already optimized run time.

students/lauraannf/lessons/lesson06/assignment/good_perf.py
```diff
@@ -0,0 +1,59 @@
...
+                if row[5][6:] == '2015':
+                    year_count["2015"] += 1
+                if row[5][6:] == '2016':
+                    year_count["2016"] += 1
+                if row[5][6:] == '2017':
+                    year_count["2017"] += 1
+                if row[5][6:] == '2018':
+                    year_count["2018"] += 1
+                if "ao" in row[6]:
+                    found += 1
```
ldconejo, line 41: Also, for cases in which you need to check multiple conditions, you can make use of the _continue_ statement to jump to the next iteration of the for loop, avoiding every single if to be checked after one has already matched:

                if "ao" in row[6]:  # Moved this one up, as it needs to be checked, regardless of whether
                                              # the year matches
                    found += 1
                if row[5][6:] == '2013':
                    year_count["2013"] += 1
                    continue
                if row[5][6:] == '2014':
                    year_count["2014"] += 1
                    continue
                if row[5][6:] == '2015':
                    year_count["2015"] += 1
                    continue
                if row[5][6:] == '2016':
                    year_count["2016"] += 1
                    continue
                if row[5][6:] == '2017':
                    year_count["2017"] += 1
                    continue
                if row[5][6:] == '2018':
                    year_count["2018"] += 1

ldconejo:
Code looks good, also good performance improvement of the previous code. 

One recommendation is to further expand your notes on why changes were made in the code. For the assignment purposes, your notes are good, but think of situations in which a code reviewer has much less background on the code you're submitting. Good notes, detailing performance improvement gains can be very helpful.

I added some comments on how you could further improve performance, but the assignment requirements have already been met.

PR #47, opened 2019-05-04
Lesson10 Complete

dfspray:


dfspray:
I couldn't find anything in the lesson files to modify with metaprogramming, so I just refactored an old version of 'linear.py.' See notes.

PR #50, opened 2019-05-07
Adding Lesson08 assignment and corresponding files

njschafi:
Adding Lesson08 assignment 'inventory.py' - along with corresponding csv files. All code graded 10/10.

### DIFF
--- a/students/njschafi/Lesson08/assignment/inventory.py
+++ b/students/njschafi/Lesson08/assignment/inventory.py
@@ -1,0 +1,39 @@
+"""Inventory Module for Lesson08 Assignment"""
+import csv
+from functools import partial
+#pylint: disable=E0602, W0613, C0103
+
+def add_furniture(invoice_file='', customer_name='', item_code='',
+                  item_description='', item_monthly_price=''):
+    """Adds furniture with given user inputs to invoice_file.csv"""
+
+    with open(invoice_file, "a+", newline='') as csvfile:
+        csvwriter = csv.writer(csvfile)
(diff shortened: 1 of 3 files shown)


students/njschafi/Lesson08/assignment/inventory.py
```diff
@@ -0,0 +1,39 @@
+"""Inventory Module for Lesson08 Assignment"""
+import csv
+from functools import partial
+#pylint: disable=E0602, W0613, C0103
+
+def add_furniture(invoice_file='', customer_name='', item_code='',
```
ldconejo, line 6: Avoid using default values for parameters, use them only when there are usage cases in which default values make sense to the code. In this case, calling _add_furniture()_ with no parameters will make the function crash:

```
>>> import inventory as inv
>>> inv.add_furniture()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ldconejo/Documents/UW/SP_Python220B_2019/SP_Python220B_2019/students/njschafi/Lesson08/assignment/inventory.py", line 10, in add_furniture
    with open(invoice_file, "a+", newline='') as csvfile:
FileNotFoundError: [Errno 2] No such file or directory: ''
```

njschafi: I updated the code so add_furniture() can handle no parameters. Updated code is in https://github.com/UWPCE-PythonCert-ClassRepos/SP_Python220B_2019/pull/60

PR #53, opened 2019-05-10
Lesson 07

LAFiorizarry:
I might be done with lesson 7, but I'm unsure if I did everything necessary. My parallel implementation doesn't really save any time, but that could be because the files aren't very big to begin with. Also, I couldn't quite figure out how to implement a lock to avoid my results being jumbled. Can you please take a look?
Thank you


students/lauraannf/lessons/lesson07/assignment/parallel.py
```diff
@@ -0,0 +1,171 @@
...
+if __name__ == "__main__":
+    start = time.clock()
+    queue = Queue()
+    queue_results = []
+    threads = [Thread(target=import_products, args=('csvfiles','inventory.csv', queue)),
+               Thread(target=import_customers, args=('csvfiles', 'customers.csv', queue)), 
+               Thread(target=import_rentals, args=('csvfiles', 'rental.csv', queue ))]
+    for thread in threads:
+        thread.start()
+        queue_results.append(queue.get())
```
ldconejo, line 168: Consider that _queue_results.append(queue.get())_ will not wait for the thread you just started with _thread.start()_ to finish.

You should follow a sequence similar to this one:

1. Launch all threads.
2. Go through the list of threads, executing a _thread.join()_ which will effectively wait for that thread to finish (other threads can continue to run without issues while this happens).
3. Pull the data from the queue.

The code will look like this:

```
    for thread in threads:
        thread.start()

    for thread in threads:
        thread.join() 
        queue_results.append(queue.get())
```

ldconejo:
Hi Laura,

See my comment on your code on a required change to ensure all your threads are completed successfully. 

Using a queue has the advantage that it has its own locking mechanism, the queue object will only allow one thread to push data into it, while other threads are made to wait.

Also, please remember to run Pylint on your code and fix any outstanding warnings.

Please let me know if you have any additional questions. Submit a new pull request once these changes are made.

Regards,

Luis

PR #60, opened 2019-05-14
Adding Lesson09 assignments + Refactored Lesson08 assignment

njschafi:
All code 10/10 on pylint.
Refactored lesson08 so it can now handle no parameters when calling add_furniture().


ldconejo:
Excellent work on assignment #9, all changes to previous assignments work as expected.

-Luis

PR #63, opened 2019-05-16
Corrected error in basic_operations.py

chelseamsmith:
capitalized DB and moved DB.init and DB.create_tables to a if __name__ == "__main__" block to ensure it won't activate when basic_operations.py is being used by the test files

### DIFF
--- a/students/chelseamsmith/Lesson03/basic_operations.py
+++ b/students/chelseamsmith/Lesson03/basic_operations.py
@@ -4,9 +4,6 @@ from peewee import *
 from customer_model import *
 
 
-db.init('customers.db')
-db.create_tables([Customer])
-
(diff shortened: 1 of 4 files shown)


ldconejo:
Hi Chelsea, 

Code looks good. Could you try to increase coverage from your unit tests? Coverage for _basic_operations.py_ is currently at 81%, trying to bring it to 90%+. The way to do this is to add coverage for the exception blocks of _update_customer_credit()_ and _delete_customer()_.

Please submit a new pull request with these changes. Keep up the good work!

-Luis

chelseamsmith:
Created
https://github.com/UWPCE-PythonCert-ClassRepos/SP_Python220B_2019/pull/65 with
improved coverage.

PR #69, opened 2019-05-20
Lesson 07

LAFiorizarry:


students/lauraannf/lessons/lesson07/assignment/parallel.py
```diff
@@ -157,15 +157,15 @@ def delete_database():
...
+               Thread(target=import_customers, args=('csvfiles', 'customers.csv', QUEUE)),
+               Thread(target=import_rentals, args=('csvfiles', 'rental.csv', QUEUE))]
+    for thread in THREADS:
         thread.start()
-        queue_results.append(queue.get())
-    end = time.clock() - start
-    print('Total Time = {}s'.format(end))
-    print(queue_results)
\ No newline at end of file
+        QUEUE_RESULTS.append(QUEUE.get())
```
ldconejo, line 168: This creates a problem, although you will probably not see it in this cause, because the data is not big enough to cause an issue. Threads should more or less follow this sequence:

1. Thread is started (and left running).
2. Next thread is started and so on.
3. Once all threads have been launched, you go over your list of threads and run _thread.join()_ for each of them. This will essentially wait for the thread in question to finish running, without impeding other running threads from continuing their run.

In this case, since you're using a queue (good idea, as the queue itself acts as a semaphore for threads trying to access it), you would start pulling the results out of the queue once all threads have finished.

The problem is not joining a thread that you have started is that it's possible for you to run the next piece of code, in your case, _QUEUE.get()_, before the thread has ended, causing your code to crash or end up with incomplete data.

Please correct this and submit a new pull request.

Also, in Python, capitalized identifiers are used for constants, QUEUE and QUEUE_RESULTS are variables.

Regards,

Luis

LAFiorizarry:
Ok, thank you. I will work on this.

Every time I run pylint, it tells me to capitalize things that are in the
if __name__ == '__main__': section. Why?

PR #75, opened 2019-05-22
Lesson7

vvinodh:


ldconejo:
Hi Vinodh,

Implementation looks good, as well as the improvement in run time. Pylint highlights three issues:

```
parallel.py:199:4: W0622: Redefining built-in '__spec__' (redefined-builtin)
parallel.py:127:15: W0703: Catching too general exception Exception (broad-except)
parallel.py:104:21: W0613: Unused argument 'directory_name' (unused-argument)
```

Please put some extra work fixing the second and third ones. The third should be quite easy, using a "catch all" for exceptions is always a big risk, as you may end up catching (and incorrectly handling) other issues you're not aware of when you created the catch all and which may require a different type of exception handling.

Keep up the good work!

-Luis

PR #81, opened 2019-05-28
Updated lesson 2 to comply with pep8 and lesson3 files

ethanenguyen:


ldconejo:
Hi Ethan,

For lesson 2, the code is good. You're currently disabling C0301 on Pylint (line-too-long). You should probably keep that one enabled and split lines that are too long, as that reduces readability of your code. On the other hand, your code still shows warnings for W1203 (passing parameters as arguments), that one is one that I'm usually ok with disabling. As with many things in Python, these _rules_ will sometimes be adapted based on the needs of individual projects.

Regards,

Luis

ldconejo:
For lesson 3, there are a couple of changes that are required. The first is that you're using _except Exception_, which will be a _catch-all_. This is fine for code you're debugging, but you should always use specific exceptions for known issues that your code needs to handle. The problem with a catch all is that your code might eventually run into an issue you were not aware of before and the catch all might handle it and inadvertently hide it. So, either add specific exceptions for known issues or remove exception handling altogether from _add_customer_.

The second thing to correct is that coverage for _basic_operations.py_ is below 90%, mostly because you're not covering the exception block. Make sure that, once you have decided how to fix exception handling, code coverage for all Python files is at 90% or higher.

Regards,

Luis

PR #88, opened 2019-06-03
Adding the lesson5 files

nubber7:
Adding the lesson5 files


ldconejo:
Hi Ramakanth,

Your implementation is good. One thing that you should change within _test_database.py_ is the multiple instances of the variable _directory_name_.  You could simplify it to use the current directory by doing this:

directory_name = "./src/data"

This will allow running your tests right away outside of your own system. You could also define _directory_name_ outside of the test class, so that you don't need to create it again inside of each class method.

Coverage is currently below 90%, I suggest you try to add exception coverage, which should push the overall code coverage past the requirement.

Regards,

Luis

PR #89, opened 2019-06-03
updated lesson_03

vmedina1014:


students/vmedina/lesson_03/basic_operations.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    try:
+        customer = Customer.get(Customer.customer_id == customer_id)
+        LOGGER.info('Found customer with id {} '.format(customer_id))
+        customer_info = {'name': customer.customer_name,
+                         'last_name': customer.customer_last_name,
+                         'email_address': customer.email_address,
+                         'phone_number': customer.phone_number}
+        return customer_info
+    except Exception as error:
```
ldconejo, line 68: Avoid using "catch all" exceptions. The risk is that your code could later run into a previously unseen issue, with that issue being intercepted by the catch all, without being handled properly.

Exception handling is better used for capturing known issues (for example, trying to enter non-numeric values into a numeric field). It's ok to use a catch all for debug purposes.

ldconejo:
See my comments on your code for some required modifications.

PR #91, opened 2019-06-05
Lesson03 Files Complete

lombardic:
Adding lesson03 files ready for review.


students/chris_lombardi/Lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,97 @@
...
+                cust_status=status,
+                cust_credit_limit=credit_limit
+                )
+            new_cust.save()
+            LOGGER.info('Added new customer %s %s', name, lastname)
+
+    except IntegrityError as error_1:
+        LOGGER.info(error_1)
+        LOGGER.info('Non-unique customer id: %s', customer_id)
+    except Exception as error_2:
```
ldconejo, line 41: Avoid using _catch all_ exception handling. The risk for released code is that you could eventually run into a corner case that trigger a different exception, which is then handled (incorrectly) by the catch all.

ldconejo:
Please remove all catch all exceptions from your code. Also, your pull request includes changes to another student's code, please make sure only files within your folder are included.

-Luis

PR #96, opened 2019-06-07
updated basic_operations for tests completion

vmedina1014:


ldconejo:
Hi Victor,

Assignment looks better. Coverage is at 85% for _basic_operations.py_, below the 90% target. Try running _$ python -m coverage html_ after a coverage run to see which parts of the code are not currently covered by your tests and add the necessary additional testing.

Regards,

Luis

PR #103, opened 2019-06-09
Incorporating lesson03 pylint comments

williamchang6:


ldconejo:
Hi Will,

It's a valid solution to add your customized _pylintrc_ file. In general, avoid full imports from large modules such as _peewee_:

`from peewee import *`

For smaller projects (like this assignment) it doesn't make a major difference, but all those imports take space in memory, which could be a much bigger concern on larger projects. Try instead to import only those elements you will need as part of your implementation.

Good luck on your next assignment,

Luis

PR #111, opened 2019-06-14
Adding lesson05 files

lombardic:
Lesson05 files complete and ready for review.


ldconejo:
Hi Chris,

Everything looks good, coverage is at 88%, below the 90% target, due to the duplicate key exceptions not being covered as part of testing. Could you go ahead and add a few tests to cover exceptions?

Thanks,

Luis

PR #137, opened 2019-07-16
assignment 3,4,5 submission

Luyao931213:
assignment4 asked me to use iterators appropriately. However, I did
not see anywhere in assignment3 that can use the iterators. Please let me 
know what I can revise. Thank you!


ldconejo:
Hi Luyao,

The requirement to add iterators, generators or comprehensions to your code is a bit open-ended, in that in might be difficult to find a place in your code when that can be added, you could use comprehension to return the list of customers from the database, for example.

Logging implementation is good, test coverage, same as for assignment 3, is low at 82%, try to add additional tests to get it to 90%+.

Regards,

Luis

PR #143, opened 2019-07-19
Completed lesson02 activity and assignment

gregdevore:
Debugging activity for the assignment is in _debugger_log.txt_
Log messages are in _charges_calc.log_


students/gregdevore/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,140 @@
...
+
+def load_rentals_file(filename):
+    """
+    Load JSON file using python json module
+    """
+    logging.debug('Loading JSON file.')
+    with open(filename) as file:
+        try:
+            data = json.load(file)
+        except Exception:
```
ldconejo, line 44: Avoid using "catch all" exception handling. The risk with this is that your code could eventually run into a corner condition that could end up being incorrectly handled (and being hidden) by the catch-all, triggering other issues. You should always add handling for specific issues you know/think your code could run into and let it crash if something else happens, so you that proper debug can be done (for some applications, this might not be acceptable, so you could have general exception handling in those cases, making sure the log shows that "something else" happen, so that the issue does not remain unnoticed).

gregdevore:
Hi Luis,

Thanks for the feedback, I really appreciate it.

I try and handle specific exceptions whenever I can, in this case I wasn’t sure what types of exceptions could be thrown by the JSON reader/writer. I looked into the documentation but didn’t find a lot of help there. Do you have any advice when you’re dealing with a new module and trying to decide which exceptions could occur and which you should try to catch? I imagine with something like the JSON reader/writer there are quite a few scenarios where it could throw an error.

Thanks again, I will make sure and always try and handle specific exceptions in future assignments.

Best,
Greg

PR #144, opened 2019-07-22
Assignment 02, first submission

RandiPet:


ldconejo:
Hi Jared,

For assignment 7, part 2, the implementation is good for 50% of the grade, you could try adding some basic functionality for adding, updating and deleting data from the database to get the full grade.

For assignment 6, part 2, unit and integration tests still have too many issues.

For assignment 4:

- Line 358: Change the donations from the first user to be a list:

    jared = Donor("Jared Mulholland", [10000])

- Line 106: _add_donor()_ should not be a class method, you're creating a instance of _DonorGroup_. The method should look like this:

```
    def add_donor(self, new_donor):
        self.donor_list_js.append(new_donor)
        self.donor_dict_js[new_donor.fullname_js] = new_donor.donations_js
        return self.donor_list_js

```
- Lines 364-366: Convert to method calls:

```
    dg.add_donor(chris)
    dg.add_donor(ben)
    dg.add_donor(kim)
```

- Line 188, _save_donors_json()_ should not be a class method. The save method should look like this:

```
    def save_donors_json(self):
        donors_json = json.dumps(self.donor_dict_js)
        with open('donor_db.json', 'w') as donor_file:
            donor_file.write(donors_json)
            print('JSON file created')
```
You still need to add some code in _load_json()_ to load the data into the actual database, so that running _create_report()_ will show the data loaded from the JSON file.

Try to finish assignment 4, hopefully tonight and I will review it right away.

Regards,

Luis

PR #149, opened 2019-07-26
Lesson 07 upload

vmedina1014:


ldconejo:
Hi Victor,

Please avoid using general exceptions, so called "catch all" exceptions, like in line 62 of _parallel.py_:

```
                try:
                    products.insert_many(product_list)
                except Exception as ex:
                    LOGGER.warning(ex)
```

The problem with doing it this way is that it will literally catch all issues, the ones you might be familiar with and the ones you haven't seen yet. The latter are the problem. In general, you want your program to crash if it runs into a new issue, so that the developer is made aware of the issue and can make a decision on how to best handle it.

Please replace all your exceptions with the specific exception type(s) (if known) or remove exception handling were no known exception exists.

Also, add some additional comments to your notes on your analysis on why parallel processing is making your code run faster.

Regards,

Luis

PR #163, opened 2019-08-02
Updated Lesson 9

vmedina1014:


ldconejo:
Hi Victor,

The code for _charges_calc.py_ is not functional. A couple of issues you need to look at:

- Line 150 has a reference to _logger_, which is not defined.
- Your decorator is coded in a way that it will try to run initialization every time, beyond that, it will try to use the first parameter passed to the function being decorated as the logging level, causing errors like this one:

`ValueError: invalid literal for int() with base 10: 'source.json'`

Also, _pngdiscover.py_ is also not finding any files beyond the first level (meaning that it does not report any .png if they are in a subdirectory. Also, I noticed you're using the _glob_ module. The idea of the assignment is for you to implement your own recursion, not to use a ready-made solution, please change this and verify that it works with several subdirectories containing .png files.

Finally, please add some additional code exercising your implementation of _database.py_ with a context manager.

Not sure why I missed these issues the first time around. The only thing I can think off is that I accidentally look at another student's submission. Please make sure you correct these issues, verify that your code is working and submit a new pull request.

-Luis

vmedina1014:
Professor,

I've updated pngdiscover.py but I haven't had the chance to update the
other python files as I had some family medical circumstances to take care
of.

I'll try my best to have that done before the end of month 4.

Thank you,

Victor A Medina

ldconejo:
Hi Victor,

_pngdiscover.py_ looks great, no major issues reported by Pylint.

I hope everything goes well with your family. At this point, your grade stands at 86, so there are no concerns with you passing the class. Your current course end date is August 31st, so there's a fair amount of time left for you to complete assignments 9 and 10. Please let me know if there's anything I can assist with.

Best regards,

Luis

PR #169, opened 2019-08-04
 Lesson 03 assignment, resubmit since I messed up the last commit :)

RandiPet:


ldconejo:
Hi Randi,

Funny title for this pull request :-) I guess a programming course if the best place to run into these kinds of issues, so that you already know how to solve them later on.

For your submission, it looks good. A few things you need to change:

- Test coverage is low at 85%, try to bring it up to 90%+, which you could achieve this by adding coverage to your exceptions.
- On line 61, you're using a general exception trap. While this is fine for debug, you should always avoid "catch all" exception handling. The problem is that they do precisely that: Catch everything. You could eventually run into a new issues that causes your code to crash; if you have general exceptions in place, they could end up capturing those new issues and obscuring them, making it difficult to figure out what's going on. You should use specific exceptions for know issues, like you do in line 47.
- Run Pylint on your code. There are a number of issues with PEP8 compliance that need to be fixed.

Happy coding,

Luis

PR #170, opened 2019-08-04
Assignment01

mzkhan87:
Dear Instructor, 

Please check my work for Assignment 1 and provide me feedback if you think I should improve on something. 

Thanks.


ldconejo:
Hi Muhammad,

Code looks good, some things to correct:

- Coverage for _main.py_ is lower than the required 90%. Try to add some additional tests to increase coverage. Also, run _python coverage html_ to get an html report showing what code is currently not being covered by your unit test.
- Running Pylint on your code shows several _too-many-arguments_ and _too-few-public-methods_ messages. I don't think you should make code changes for this, but instead use Pylint's feature to selectively disable some warnings / rules.

To disable a warning in Pylint, add a comment like this at the beginning of your code:

_# pylint: disable=Rxxxx_ where _Rxxxx_ would be the rule to disable (for example: R0903).

Once you have corrected these, please submit a new pull request.

Regards,

Luis

mzkhan87:
Hi Luis,

Actually, I suppressed the *too-many-arguments* and *too-few-public-methods
messages using the .pylintrc file. Per your feedback, I have added the
statement*
*#pylint: disable = R0903, R0913*
*It was nice to learn we could use the message code like Rxxxx. I had only
known before to use statement like*

*#pylint: disable = too-many-arguments, too-few-public-methods*

*I have resubmitted the assignment. *

*Thanks so much! *

PR #185, opened 2019-08-10
Added Lesson 7 assignment

chelseamsmith:


ldconejo:
Hi Chelsea,

The current implementation has a couple of issues:

1. You're not joining the thread, meaning that you run _my_queue.get()_ without knowing whether the thread that was supposed to put the data in the queue finished or not. For that, you should run _threadX.join()_ where the _X_ would be replaced by 1, 2 or 3, respectively.

2. You should collect the results from the queue at the end, after all three threads have launched and finished.

The sequence should go like this:

1. Launch thread 1.
2. Launch thread 2.
3. Launch thread 3.
4. Join thread 1 (this will not prevent any of the threads from running).
5. Join thread 2 (same as above).
6. Join thread 3 (same as above).
7. Collect the data from the queue.

Check the execution time and, if appropriate, revise your conclusions.

Regards,

Luis

chelseamsmith:
Corrected lesson 7:
https://github.com/UWPCE-PythonCert-ClassRepos/SP_Python220B_2019/pull/189

PR #193, opened 2019-08-13
Lesson9 assignment

chelseamsmith:


ldconejo:
Hi Chelsea,

Everything looks good. For _database.py_ you're using catch all exception handling, please replace those with specific exceptions or remove them altogether. Using a bare _except_ is fine for debug purposes, when you need the code to run to completion no matter what, but they're a bad idea for longterm support, as they would catch and handle issues you're aware of, and new issues (for which different handling might be required), which is not desirable.

Regards,

Luis

PR #195, opened 2019-08-14
assignment lesson03 : database interactions with peewee library

JoeNunnelley:
included here is the following:
- basic_operations - with the functionality of the assignment
- __init__ - file for the modules
- pylintrc - for the project
- test.sh - for running the testing, coverage and linting operations. Please note that this script and running the .py lib itself will create a db file but also remove it upon completion so you may not see the .db file.  To see it, comment out the deletion command in the script
- test_integration - an integration test for the module
- test_unit - unit tests for the module


ldconejo:
The fact that you create your own _test.sh_ makes reviewing your submissions very easy, thank you, Joe. Good implementation and good test coverage as well. Keep up the good work.

-Luis

PR #200, opened 2019-08-18
Adding lesson04 assignment

gregdevore:
Just wanted to note that I don't know if my implementation of 'return_customer_ids()' and 'print_all_customers()' illustrates the best use of a generator. My thought was that for a large customer database, you wouldn't need to store all of the customer information before printing it, you would just yield one customer at a time and print their data.


ldconejo:
Hi Greg,

Good job on this one. On the use of the generator, the assignment is a bit tricky in that it's a bit difficult to force the use of elements such as generators. I look forward to your next assignment.

Regards,

Luis

PR #202, opened 2019-08-19
rfelts assignment 4

rfelts:
I didn't see a way to use iteration with any of the current functions I developed in lesson 03. So I added a couple new methods to show iteration.


ldconejo:
Hi Russel,

Code looks good. This assignment is a bit tricky, in that there are not many obvious places in the code to add the new structures. Your solution was a good one.

Regards,

Luis

PR #221, opened 2019-09-09
Assignment05

mzkhan87:
Please review the assignment 05 and provide your feedback for improvement if needed. 
Thanks


ldconejo:
Hi Muhammad,

Your implementation looks good. One thing to change is the use of _catch all_ exception handling, meaning using _except_ without specifying an exception. The problem with this is that it will catch (and handle) ANY exceptions, including those caused by issues that you might not be aware of while writing your code.

Instead, replace them with exception handling for specific exception (for example, _KeyError_) or remove the try/except blocks altogether. In general, using a bare _except_ should be reserved for cases in which you are debugging your code and never for final code.

Let me know if you have any questions and keep up the good work.

Regards,

Luis

PR #229, opened 2019-09-15
Lesson01

ericae14:


students/erica_edwards/lesson01/assignment/test_integration.py
```diff
@@ -0,0 +1,81 @@
+import sys
+import os
+sys.path.insert(1, f"{os.getcwd()}\\inventory_management")
```
ldconejo, line 3: This works well, except if the user (or the instructor ;-) ) is using Linux or MacOS, as the folder divider is the slash instead of the backslash. One way to prevent this is to have your code determine which platform type you're running with is to use _os.name_, which will return _posix_ for Unix, Linux or MacOS (which would cover all slash users).

ericae14:
You mean something like this?
I had lots of trouble getting my imports straightened out so that I could both run the application and run the tests.
When I got them to work for the tests then they were incorrectly imported if I wanted to actually use the application.
It would be nice to have some additional information on importing. Do you have any suggestions for me to look up?

import sys
import os
if os.name.lower() == "nt":
    delimiter = "\\"
else:
    delimiter = "/"
sys.path.insert(1, f"{os.getcwd()}{delimiter}inventory_management")

ldconejo:
Hi Erica,
 
Yes, it can be a simple thing. _os.name_ will return _posix_ (all lowercase) for MacOS, Linux and Unix, all three of which use slash ('/') so you could do something like this:

```
import sys
import os
if os.name.lower() == "nt":
    sys.path.insert(1, f"{os.getcwd()}\\inventory_management")
else:
    sys.path.insert(1, f"{os.getcwd()}/inventory_management")
```

The ultimate reference to Python imports is the official documentation:

https://docs.python.org/3/reference/import.html 

It's a bit of a dense read, but it does cover everything, from priorities in the path to relative imports.

Regards,

Luis

PR #230, opened 2019-09-15
Lesson01

pkoleyni:
Hi Luis 
I have created test.py and the highest coverage i could get is 92%. 
The following lines are not getting tested by my test.py file, i need help to find out why and update test.py

```

`def item_info(): 

    """Item Info""" 

    item_code = input("Enter item code: ") 

    if item_code in FULL_INVENTORY: 

        print_dict = FULL_INVENTORY[item_code] 

        for key, value in print_dict.items(): 

            print("{}:{}".format(key, value)) 
    else: 
        print("Item not found in inventory") 

 

def exit_program(): 
    """Exit program""" 
    sys.exit() 

if __name__ == '__main__': 
    FULL_INVENTORY = {} 
    while True: 
        print(FULL_INVENTORY) 
        main_menu()() 
        input("Press Enter to continue...........") `
```

```
inventory_management/__init__.py                        0      0   100%
inventory_management/electric_appliances_class.py      15      0   100%
inventory_management/furniture_class.py                15      0   100%
inventory_management/inventory_class.py                13      0   100%
inventory_management/main.py                           52      8    85%
inventory_management/market_prices.py                   2      0   100%
-----------------------------------------------------------------------
TOTAL                                                  97      8    92%
```

### DIFF
--- a/students/pooria_k/lesson01/test.py
+++ b/students/pooria_k/lesson01/test.py
@@ -1,0 +1,79 @@
+#!/usr/bin/python
+from unittest.mock import patch
+from unittest import TestCase
+from inventory_management.inventory_class import Inventory
+from inventory_management.electric_appliances_class import ElectricAppliances
+from inventory_management.furniture_class import Furniture
+from inventory_management.main import *
+
+class InventoryTest(TestCase):
+
+    def setUp(self):
+        self.inventory_1 = Inventory('1234', 'lamp', '45', '10')
+        self.inventory_1_dic = self.inventory_1.return_as_dictionary()
+        self.electric_appliances_1 =ElectricAppliances('4455', 'TV', '2200', '140','Samsung', '120')
+        self.electric_appliances_1_dic = self.electric_appliances_1.return_as_dictionary()
+        self.furniture_1 = Furniture('2020', 'sofa', '1200', '100', 'leather', '22x33x44')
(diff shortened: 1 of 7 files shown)


students/pooria_k/lesson01/test.py
```diff
@@ -0,0 +1,146 @@
...
+        with self.assertRaises(SystemExit):
+            main.exit_program()
+
+    def test_item_info(self):
+
+        with patch ('builtins.input', side_effect = '444'):
+            main.FULL_INVENTORY={}
+            self.assertEqual(main.item_info(),  print("Item not found in inventory"))
+
+        with patch ('builtins.input', side_effect = '444'):
```
ldconejo, line 100: This line should be: 

`with patch ('builtins.input', side_effect = ['444']):`

Note the additional square brackets ([]). _side_effect_ is treated as an iterable, if you just leave the string, each time you call _input_ from _main.item_info()_ it will really pull one character (a '4', in your case). 

This should bring coverage for main.py to 90%, which meets the requirement. There are ways to test the main execution block of the file, but that's usually not done.

Regards,

Luis

ldconejo:
Hi Pooria,

I commented your code with a small change you need to make, that should bring your coverage up to 90%.

Regards,

Luis

PR #232, opened 2019-09-19
Submitting Lesson 01 Revisions

cholso:
1. Increased coverage in inventory_management\main.py from 82% to 90%
2. Fixed test_integration.py so it works now.


ldconejo:
Good job, coverage is indeed at 90% now and the integration test works as well. I look forward to your next submission!

-Luis

PR #234, opened 2019-09-19
Pull request to help finish up lesson 1 - fixing import statements

JonCracolici:
As we discussed I'm more or less there, but I'm having trouble with my import statements. Looking forward to your feedback!


students/jon_cracolici/lesson01/assignment/test_unit.py
```diff
@@ -0,0 +1,221 @@
+"""Unit Test Module for Inventory Management Software."""
+
+import io
+import os
+import sys
+DIR_PATH = os.path.dirname(os.path.realpath(__file__))
```
ldconejo, line 6: This one should instead be:

```
DIR_PATH = os.path.dirname(os.path.realpath(__file__)) + "/inventory_management"
sys.path.append(DIR_PATH)
```

This will allow the other classes to import directly from _inventory_management_ without having to add _inventory_management_ to the import statement.

ldconejo:
Hi Jon,

I just commented on your _test_unit.py_, there's a small change that needs to be made. The same change applies to _test_integration.py_. You do not need the _DIR_PATH_ and the _sys.path.append()_  lines anywhere else. 

Once those lines are removed from the files inside _inventory_management_ your code should is graded at 10/10 by Pylint.

Let me know if this makes sense, if you continue to have issues, submit another pull request with your latest version of your code for me to take look.

Happy coding,

Luis

PR #240, opened 2019-09-23
Help on assignment 1

gnar5tar:
Pull request for help on assignment 1 from our email discussion. Thanks!


students/zach_thomson/lesson01/inventory_management/electric_appliances_class.py
```diff
@@ -0,0 +1,21 @@
+'''module that creates an electric appliances class'''
+
+# Electric appliances class
+from inventory_management.inventory_class import Inventory
```
ldconejo, line 4: Change this line back to just calling _inventory_class_, same for _furniture_class.py_:

```
from inventory_class import Inventory
```

students/zach_thomson/lesson01/test_unit.py
```diff
@@ -0,0 +1,52 @@
+'''unit testing individual components of inventory management'''
+
+from unittest import TestCase
+#from unittest.mock import MagicMock
```
ldconejo, line 4: In here, at the _inventory_management_ directory to the Python PATH, by doing the following:

```
import sys
sys.path.append("./inventory_management")
```

Note I'm using a UNIX-path, since I'm using a Mac, it might be a bit different if you're on a PC.

ldconejo:
Hi Zach,

I added a couple of comments on changes you can make to get your unit testing running without having to break _main.py_ and associated modules. In general, you don't want your unit testing forcing changes into the code being tested. I look forward to your next pull request!

Regards,

Luis

students/zach_thomson/lesson01/test_unit.py
gnar5tar: That solved my issue - thanks so much for the help!

PR #241, opened 2019-09-23
Lesson 07 Assignment files

rfelts:
Lesson 07 assignment files for rfelts.


ldconejo:
Interesting results. _ThreadPoolExecutor_ is mostly a simplified implementation, intended to remove a lot of the learning curve. In any case, on my system, the linear implementation is slower than the other two.

-Luis

PR #242, opened 2019-09-24
Assignment 1

gnar5tar:
Thanks for the help on getting my test_unit.py to work with the modules. 

Pylint is at a 10. I only disabled too-many-arguments, too-few-public-methods, and invalid-name (for the 'for k, v in print_dict.items():' portion since that seems like fairly accepted syntax). 

My test_unit.py has 93% coverage. 

I would appreciate any feedback, especially on the test_integration. I mainly tested creating an inventory dictionary and then probing it for various included items. Is this enough/correct?

Thanks!


ldconejo:
Hi Zach,

Implementation looks good, I appreciate you including details on what warnings you disabled from Pylint. The ones selected are ok.

On the integration test, the key is to have comprehensive coverage for the unit test and one or more actual usage cases in the integration test (for an real-life project, finding the right combinations for integration testing can be a bit tricky, in my opinion). One thing I try on occasion, in particular for a smaller project, is to get my integration test to also get to 90%+ coverage of the code, but that's not always possible.

Remember to also submit your code on Start Learning (open EdX), as that's were I grade it. Send me an email once you have made the submission. Good luck with assignment 2!

-Luis

PR #244, opened 2019-09-25
Assignment 09, first submit

RandiPet:


ldconejo:
Hi Randi,

Did you try running _charges_calc.py_ after implementing your decorator for logging? Currently, I get the following error:

```
$ python charges_calc.py -i source.json -o output.json -d 3
source.json
Traceback (most recent call last):
  File "charges_calc.py", line 130, in <module>
    DATA_INPUT = load_rentals_file(ARGS.debug, ARGS.input)
TypeError: 'NoneType' object is not callable
```

Note that your current implementation of _log_decorator_ doesn't return the function being decorated, since _return make_logger_ has been nested within _make_logger_. Anyway, take a look and let me know.

-Luis

ldconejo:
Also, _png_discover.py_ is empty, looks like the corresponding code was not added in the pull request.

ldconejo:
For _database.py_, please include some code that showcases your implementation of context managers.

PR #245, opened 2019-09-26
Lesson02

ericae14:
In pdb I discovered the script was closing due to a value error. The error was caused by value total days being negative. To fix the issue with negative total days I added a new list to track the bad rental keys. I then added an if statement to check for total days that are less than zero. If the total days are less than zero. The record is rejected and the rental id is added to the list bad_data_keys . After the try/except block has finished creating the additional fields the bad records are removed from the data that will be saved to output.json. I also updated the except statement to log errors. The rental id for the record that caused is then added to the list bad_data_keys. It will then be deleted after the try/except block is complete.


ldconejo:
Hi Erica,

Implementation looks good. A couple of things to change:

1. Your code run without any arguments, it should instead error out, telling the users which parameters are required / optional.
2. _init_logger()_ takes one parameter, but doesn't use it (it calls args.debug directly instead), please modify so that _args.debug_ is passed to the function (and used).
3. Run Pylint on your code and correct the issues detected until you get 10 out 10 grade (currently, your code is at 3.85/10). Note that Pylint allows you to use comments in code to disable some warnings. If you decide to use this feature, please indicate which warnings you're disabling and the reason.

Also, remember to submit your assignment on open edX for grading.

Regards,

Luis

ericae14:
Hi Luis,

I will work on correcting the issues you mentioned below. I am however a bit concerned about my assignment being saved in edx  as I did submit it soon after I put my assignment on git. I will try to resubmit the assignment after I make the changes again to see if you see it.

Thanks,

Erica

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

ericae14:
Hi Luis,

I did a pull request to with my corrections for lesson 2 but the pull request for lesson 3 got merged into also. Not sure how to fix it.

Thanks,

Erica

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

PR #247, opened 2019-09-26
Assignment09, resubmit to correct files

RandiPet:


ldconejo:
Hi Randi,

Looks good, both _charges_calc.py_ and _pngdiscover.py_ are working now. Please also add the code request to exercise your changes to _database.py_, it could be done either with a _main_ block within the same file or if you want as a separate unit_test. 

Also, please make sure you run Pylint on all three files and make sure you address any outstanding coding issues (there's a bare exception there, make sure you either remove the _try/except_ block or update it to include a specific exception type(s)).

Regards,

Luis

PR #250, opened 2019-09-28
Adding lesson09

gregdevore:


ldconejo:
Hi Greg,

Could you submit the assignment on open edX, so that I can grade it?

Thanks,

Luis

gregdevore:
Hi Luis,

I just submitted it, sorry for the delay! I tried to submit it earlier and got a weird error. I just got back from dinner and it appears to be working now. Let me know if you need anything else!

Thanks,
Greg

PR #251, opened 2019-09-28
lesson01 activity and assignment

brianminsk:
completed lesson01 activity and assignment


ldconejo:
Hi Brian,

Looks good. From a formal standpoint, _main.py_ is missing the module docstring (these are useful for someone looking at your code to quickly figure out what each file is doing). Keep up the good work.

-Luis

brianminsk:
Ah - that was an oversight on my part. Thanks.

Brian

PR #255, opened 2019-09-30
Assignment 2 Submission

gnar5tar:
This pull request is for assignment 2. 

output.json is the final output from the edited charges_calc.py

Will submit on the course website as well. 

Thanks!


ldconejo:
Hi Zach,

Good job on this one. One thing to consider is that you could make a parameter such as _-debug_ optional with a default value when not provided, but that's your choice as developer. Looking forward to your next assignment submission.

Regards,

Luis

PR #256, opened 2019-09-30
Lesson02 Activity and Assignment

brianminsk:
Final files for the Lesson02 activity and assignment

Regarding the debugging for the assignment:
I stepped through the program using pdb and discovered that in the first iteration through the data there was a problem with the data - rental_end was before rental_start. This cascaded to a point where it was trying to take the square root of a negative number when calculating sqrt_total_price. You can see how I tracked in in my pdb output.

The rest of the debugging was based on running the script with logging to a file. When there was an exception, it usually corresponded to bad data, so I looked in the log file to see what the data actually was and confirmed with the record in the input json file. As I wrote code to handle the problems and exceptions I iterated on this method of identifying and coding to handle the problems.


ldconejo:
Hi Brian,

Thank you for commenting on your debug process on your pull request, it makes it easier to review your results.

-Luis

ldconejo:
Hi Brian,

The following command line crashes your code:

```
$ python3 charges_calc.py -i source.json -o test.json
```

One option is to modify your command line arguments to make _-debug_ a mandatory parameter or give it a default value.

brianminsk:
Hi Luis,

Thanks for pointing that out - I should have tested it without passing a
command line -debug parameter.

It's interesting though, because I had set up a default value ("0") for it
in the argument list for the function I created to set up the logger and
had assumed that if None was passed to it for the logging level it would
assign the default value. So I learned this morning that None will stay
None regardless. Good learning experience that makes sense now that I think
about it. haha

I am not sure if you are asking me to resubmit with a correction to that?
Regardless, I will make the correction and make a new pull request.

Brian

PR #260, opened 2019-10-03
lesson 3 files

franjaku:


students/franjaku/lesson03/src/basic_operations.py
```diff
@@ -0,0 +1,107 @@
...
+                                           name=name,
+                                           last_name=last_name,
+                                           home_address=home_address,
+                                           phone_number=phone_number,
+                                           email_address=email_address,
+                                           status=status,
+                                           credit_limit=credit_limit)
+            new_customer.save()
+            LOGGER.info('New Customer: %s, created and saved to the database.', customer_id)
+        except Exception as err:
```
ldconejo, line 40: Avoid using general, "catch all" exceptions. Instead, use only try / except blocks that specify the specific exceptions to be handled by the _except_ portion. This will cause any _new_ exceptions to crash your program, letting you know that previously not seen exception type has come up.

ldconejo:
Hi Francisco,

Code looks good, see my comments on replacing general exception blocks. Please submit a new pull request once this change is made.

Regards,

Luis

PR #261, opened 2019-10-04
Added Some Assignment01 Files

ajaggarcia10:
Hello,

I've been working through assignment 01 and there are a couple of things I stumbled upon. 
For Pylint, you stated to use the pylintrc file for the class, and looking at the discussions it seems like we are just using the original config file. When I use Pylint, I don't get a code of 10 for me, but I want to assume its ok.  I get the 'too few public methods', which I know we can partially ignore from the lesson videos. I also get 'too many arguments', which looking at the configs, seems to be something that I can change, but not sure if I should change it within the configs. I also get 'similar lines', which I also feel like can be ignored. So my question is if its okay to ignore these errors, and change that within the configs to get a score of 10 or should I just leave the configs and assume the code is ok?

When I get to writing the test_unit file, I tested the other modules, but I'm having trouble figuring out how to go about testing main.py. I understand that we need to Mock things, but I'm not sure where to begin. I'm seeing the usage of patch in other's work, but I don't fully understand how it works and can't find extra helpful information about it and I'm not sure about the other ways to go around testing main.py.


ldconejo:
Hi Andrew,

First of all, welcome to the class. To the Pylint question, you will find situations in which a particular warning really does not apply to your project. As you mentioned, one option is to disable those warnings using .pylintrc (which you need to include on your pull request) or disable individual warnings by using comments on your code. For example, if you want to disable _too-many-arguments_, you'd have to add this comment to _furniture_class.py_:

```
# pylint: disable=R0913
```

This will allow targeted disabling on individual files (and will not require you to include .pylintrc). It's ok to setup your code to ignore those errors, I'll let you know if you disable a warning that I think should be kept enabled (for example, Pylint will warn you against using a general, catch-all exception, which is not a good coding practice).

On using mock for unit testing, I'll send you some sample files for you to look at. The idea is that you should be able to test things such as user input by mocking said input. Patching is another option (not covered in the sample files, those are a simpler implementation). Let me know if you'd like to schedule a quick Office Hours session to go over the topic, that's sometimes a better option.

Regards,

Luis

PR #263, opened 2019-10-06
Complete Assignment 2:

bryancoates:
- Added argument for logging level and function to handle
 - Added debug, error, and warning logs
 - Added exception for start date later than end date
 - Added exception for missing date values
 - Cleaned up formatting to pass pylint


ldconejo:
Hi Bryan,

Looks good. One thing to change is that your code should reject invalid debug levels (right now, using 0, 4, 5, etc. will work and setup logging as if they were valid values. Please submit a new pull request once this is corrected.

The warning you disabled on Pylint is ok. Thank you for indicating the reason to disable it.

Regards,

Luis

bryancoates:
Hi Luis,

Thank you for the feedback!  I've corrected the issue and created a new pull request.

Thanks,

Bryan

PR #267, opened 2019-10-08
Adding in Lesson03

cholso:
Submitting Lesson 3 Assignment

### DIFF
--- a/students/chelsea_nayan/lesson03/basic_operations.py
+++ b/students/chelsea_nayan/lesson03/basic_operations.py
@@ -1,0 +1,87 @@
+'''Basic Operations for the HP Norton database'''
+
+# pylint: disable=too-many-arguments
+# pylint: disable=wildcard-import
+# pylint: unused-wildcard-import
+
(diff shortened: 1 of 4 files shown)


ldconejo:
Hi Chelsea,

It looks generally good. A couple of things to correct:

1. It should be possible to run _test_unit.py_ and have it create and populate a database that will suit your testing needs. Right now, unit testing fails (missing tables). It goes a little further after running _create_db.py_, but still fails. Make sure you get all your unit tests to pass (and 90%+ coverage) by running only _test_unit.py_.

2. It's ok to do a wildcard import here (I see you disable the associated Pylint warnings). Just keep in mind that, for code intended for release, it's causing for code to be loaded that will not be used, which might become a performance issue in bigger projects where several modules need to be loaded.

Keep up the good work!

-Luis

ldconejo:
Also, your Pylint disables on _basic_operations.py_ are not fully implemented. Make sure you run Pylint ahead of submitting your update.

PR #268, opened 2019-10-08
Assignment08

mzkhan87:
Dear Instructor, 

Please review my assignment 8. 

Thanks.


ldconejo:
Hi Muhammad,

Implementation looks good. I assume you're running Pylint with a .pylintrc file that has E1121 disabled (which is ok). Next time, make sure you include that file with your submission or add a Pylint disable comment to your code, like this one:

```
# pylint: disable=E1121
```

-Luis

PR #269, opened 2019-10-08
Added Assignment01 Files

ajaggarcia10:
Hey Luis, 

Thanks for the sample mock files, they were really helpful in helping me grasp the idea of mocking better so I could get through the assignment.

I think the only issue I have right now is that I'm only at 85% coverage for main.py, but I'm not sure how to test the extra lines to increase the coverage.


ldconejo:
Hi Andrew,

I'm glad the code I sent helped! For coverage purposes, don't worry about the _main_ block. There are ways to unit test it but it's usually not a common practice, as the idea of the test the individual methods within a module. Now, for the _item_info()_ method, you do need to cover the case in which the product does exist. For assignment 1, just make sure you trigger that code (by mocking input to a product that exists). Later, when you see context managers, we can talk about methods to do unit testing with stdoutput (i.e., things that get printed on screen).

Once you add coverage for _item_info()_ simply submit a new pull request, I'll regrade your assignment on open edX without you having to resubmit there.

Regards,

Luis

PR #276, opened 2019-10-13
Updating Lesson03

cholso:
Updated so unittest shouldn't fail, but coverage is a bit wonky (unsure how to fix).


students/chelsea_nayan/lesson03/tests/test_unit.py
```diff
@@ -0,0 +1,178 @@
...
+
+        LOGGER.info('Finished testing add_customer function!')
+
+    def test_search_customer(self):
+        '''Testing search_customer function'''
+
+        LOGGER.info('Testing search_customer function!')
+        setup()
+
+        add_customer(*self.CUSTOMER_LIST[3]) # Danny Devito
```
ldconejo, line 92: This test is not testing for _search_customer()_ but for _add_customer()_, so your coverage is reduced there.

ldconejo:
Hi Chelsea,

A few things to correct:

1. _test_unit.py_ is not testing _search_customer()_ (see my comment on your code).
2. You need to add testing for exceptions.  Here's some sample code you can adapt to cover the different exceptions in your code:

```
    def test1(self):
        setup()
        with self.assertRaises(IndexError):
            search_customer(*self.CUSTOMER_LIST[10])
```

Here, we call _assertRaises_ within a context manager, so that triggering the exception will not crash the code. You'll need to make sure _assertRaises_ receives the correct exception as a parameter. 

Regards,

Luis

PR #286, opened 2019-10-20
Including .pylintrc file to suppress warnings for pylintr

KRTstudent:
My coverage report says 86% for main.py and it was suggested i cover item_info(). My original test_unit covers it in lines 151-169. What's going on?

### DIFF
--- a/students/kevin_t/Lesson1/assignment/.pylintrc
+++ b/students/kevin_t/Lesson1/assignment/.pylintrc
@@ -1,0 +1,586 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-whitelist=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
+
+# Add files or directories matching the regex patterns to the blacklist. The
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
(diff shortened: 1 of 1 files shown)


ldconejo:
Hi Kevin,

There are a couple of issues with your unit testing. One is this one:

```
        with patch('builtins.input', side_effect='1000'):
            self.assertEqual(item_info(), print(example_failed_string))
```

In this case, you're really not comparing strings, but the return of both functions. Both _item_info()_ and _print()_ return _None_. You can verify this by replacing the assertion with:

```
        with patch('builtins.input', side_effect='1000'):
            self.assertEqual(item_info(), None)
```
Or even with:

```
        with patch('builtins.input', side_effect='1000'):
            self.assertEqual(item_info(), print("hello world"))
```
The other issue is also responsible for your coverage being below 90%. The _side_effect_ parameter in _patch()_ treats the value you give to it as a list of elements. Since you're passing a string ('1000'), it actually uses it as _(1,0,0,0)_. This doesn't affect the test, as it will try to search for item "1", which doesn't exist anyway.

However, when you search for item "45", that one fails, because when treated as a list, it becomes _(4,5)_, and item "4" doesn't exist.

So, the assertion on line 162 of _test_unit.py_ should look like this:

```
        with patch('builtins.input', side_effect=('45',)):
            self.assertEqual(item_info(), print(example_string))
```

Try that, your coverage for _main.py_ should go to 92%.

-Luis

P.S.: There are ways to capture the standard output (i.e., the result of a _print()_ call) for the purposes of unit testing, but we're not covering those here as their usefulness is more debatable, as you will normally not test a print statement directly.

ldconejo:
Your custom _.pylintrc_ looks good.

PR #287, opened 2019-10-21
Assignment 5 submission

gnar5tar:
Pull request for mongoDB assignment. Will submit on course website as well - thanks!


ldconejo:
Hi Zach,

Everything looks good. One minor recommendation would be to include some code in _test_database.py_ to clear all collections before testing. Not having that created an interesting issue on my side, as a previous review of another student's assignment left entries which used _customer_id_ instead of _user_id_, causing your test to crash on my local mongo server. Just something to consider, as in general you don't want the results of your testing to depend on a database status you don't control.

There could be exceptions, of course, you might want to have a pre-built test database what thousands of records, which you don't want to recreate each time you're running a test.

Keep up the good work, looking forward to your submission for assignment #6.  

-Luis

PR #288, opened 2019-10-22
Modified test_item_info(self): to correctly test the item that exists…

KRTstudent:
… and does not exist, bringing coverage above 90%


KRTstudent:
Luis,

I pushed my lesson 2 homework before it got graded so it all went in with
the same pull request. Are you able to grade my lesson 2 submission? I
don't see a grade for it currently

ldconejo:
Hi Kevin,

Sure. I just checked on the course website, but there's no submission from you for assignment 2. Could you add that submission? Then I can grade your assignment.

Regards,

Luis

KRTstudent:
Ah, I had forgotten about the course website. I have completed that now.
However, my pull request was completed so you should be able to see the
homework files submitted. Please let me know if this is not the case.

[image: image.png]

ldconejo:
Hi Kevin,

Looks good. One small change: The debug parameter should not accept invalid values. Right now, if I enter -debug 10 it will still work and create an output log. Could you correct that?

Thanks,

Luis

PR #290, opened 2019-10-25
Completed Lesson 4

bryancoates:


students/bcoates/lesson04/basic_operations.py
```diff
@@ -0,0 +1,104 @@
...
+        logging.debug(f"Unable to find customer with id:  {customer_id}")
+        raise DoesNotExist
+
+def list_active_customers():
+    """ Return count of active customers in the database """
+
+    active_members = Customer.select().where(Customer.status == 'active').count()
+
+    # Alternate solution for comprehension / generator exercise
+    i = 0
```
ldconejo, line 94: You're not using this variable, so might as well remove it.

PR #292, opened 2019-10-26
Anthony McKeever - Lesson 2 Content Submission

Snip3rM00n:
This is my submission for all lesson 2 content including the activity and assignment.

For the assignment I encountered two bugs through multiple debug sessions:
1. The application would silently fail if the range between rental_start and rental_end was invalid resulting in a negative range.
2. The application would crash if one of the date values was empty.

The first bug fix involved implementing a check for if rental_end was less than rental_start and log if this were to happen.  Upon that, continue to parse the information and log the error.  Eventually, I decided to not let the error occur and instead log the value object into a Debug statement.

The second bug fix involved refactoring the try/except block to capture when any of the items in the value object caused a problem.  Upon entering the except block, we would log the exception as Error and log the value into a Debug statement.

**_Squashed Commit History:_**

* Initial commit of lesson 2 files

* Finish assignment

Fix Flake8 and pylinting issues
Resolve bugs in charges_calc.py
Add Logging to charges_calc.py

* Add Activity

Add activity content
Update readme
Move Assignment 1 out of unnecessary code folder

* Add headers


ldconejo:
Hi Anthony,

Good work on this one. Your debug documentation was also very useful in understanding your thought process. 

-Luis

PR #294, opened 2019-10-27
Completed lesson 5

bryancoates:


ldconejo:
Nice touch in implementing the connection to the database as a context manager, as you will run into that as a requirement for a future assignment.

PR #308, opened 2019-11-09
Updating Lesson03 and Lesson04

cholso:
Apologies about the many major errors, its been a long month. I made suuuuure that unit testing throws no errors, all tests pass, and coverage is 90% or above.


ldconejo:
Hi Chelsea,

Looks great! Logging on lesson 4 also works as expected. Both assignment get a negative grade on Pylint, something that should be a very quick fix: Pylint is complaining about unused imports, caused by wildcard imports. You did add code to disable that warning, but it's missing the _disable_ keyword:

You have:

`# pylint: unused-wildcard-import`

It should be:

`# pylint: disable=unused-wildcard-import
`
Make sure you fix it in your code, I won't subtract any points because of it. Looking forward to your submission for assignment 5.

-Luis

PR #309, opened 2019-11-10
Review of lesson 4

ericae14:
I didn't find much to do in this lesson. There were no loops in my basic operations file so no place for generators, iterators, or comprehensions. Did I misunderstand the assignment?


ldconejo:
Hi Erica,

This assignment is a bit open ended, in that as you pointed out, there aren't always clear locations to use any of those structures. Logging for the database assignment was implemented correctly.

Regards,

Luis

PR #312, opened 2019-11-11
Added Rough Draft for Assignment02

ajaggarcia10:
I think I got most of what I needed for this assignment,  I'm more curious as to if I'm going in the right direction with my code, and then I'll go back and clean it up/add more code if I missed anything. Thanks!


ldconejo:
Hi Andrew,

Looks pretty good. A couple of changes:

- Invalid logging levels: The code should error out if an invalid debug level is entered (for example, 10).
- For debug level 0, no log file should be created.
- Make sure you run Pylint on your code and fix any outstanding issues. You an use selective disabling of warnings (sparsely, in cases where it makes sense).

Regards,

Luis

PR #313, opened 2019-11-11
Lesson06

pkoleyni:


students/pooria_k/Lesson06/good_perf.py
```diff
@@ -0,0 +1,46 @@
...
+        "2015": 0,
+        "2016": 0,
+        "2017": 0,
+        "2018": 0
+    }
+    found = 0
+    start = datetime.datetime.now()
+    with open(filename) as csvfile:
+        reader = csv.reader(csvfile, delimiter=',', quotechar='"')
+        new_ones = []
```
ldconejo, line 26: _new_ones_ variable is initialized but not used. You should be able to remove it.

PR #316, opened 2019-11-12
Lesson 4 homework

KRTstudent:
Homework for week 4 submission

### DIFF
--- a/students/kevin_t/Lesson4/basic_operations.py
+++ b/students/kevin_t/Lesson4/basic_operations.py
@@ -1,0 +1,105 @@
+"""
+Store customer data from HP Norton in a relational database (sqlite3).
+"""
+import logging
+from peewee import SqliteDatabase, Model, IntegerField, CharField, BooleanField, DecimalField
+
+database = SqliteDatabase('customer.db')
+database.connect()
+database.execute_sql('Pragma foreign_keys = ON;')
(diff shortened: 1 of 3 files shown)


students/kevin_t/Lesson4/basic_operations.py
```diff
@@ -0,0 +1,105 @@
...
+    try:
+        with database.transaction():
+            new_customer = Customer.create(customer_id=customer_id, first_name=first_name,
+                                           last_name=last_name, home_address=home_address,
+                                           phone_number=phone_number, email_address=email_address,
+                                           status=status, credit_limit=credit_limit)
+            new_customer.save()
+            LOGGER.info(f'Database add successful {first_name} {last_name}')
+
+    except Exception as error_info:
```
ldconejo, line 47: Avoid using a _catch all_ exception. The risk with these is that they will catch (and handle) ANY exceptions, expected and unexpected, which might cause new issues to be hidden away, complicating the debug process. Instead, use specific exceptions that you know could occur or remove your _try/except_ block altogether.

ldconejo:
Hi Kevin,

The assignment requires that a log file named _db.log_ is created by your code. Please modify your implementation so that this file is created and log messages are printed both there and on screen.

Thanks,

Luis

PR #319, opened 2019-11-15
Edited Assignment02

ajaggarcia10:
As requested, I checked the file with Pylint and made it so it had a score of 10, I also added an exception when an incorrect debug level is inputted.

One issue I had was not having a text document made if the debug level was zero. I tried using logging.disabled = True to disable the file_logging, which did not work for me, so I tried using logging.removeHandler, which also did not seem to do anything, so I ended up moving the fileHandler creation to each individual debug level, except zero, which did solve the issue, but does result in a lot of repeated code, but I'm not exactly sure how else to prevent a file creation at debug level 0.


students/andrew_garcia/lesson02-assignment/charges_calc.py
```diff
@@ -0,0 +1,118 @@
...
+
+import argparse
+import json
+import datetime
+import math
+import logging
+
+
+def debug_setup(debug_level):
+    """ Sets the logging file name and levels """
```
ldconejo, line 13: One alternative is to add these two lines here:

```
if debug_level == 0:
    logger.disabled = True
    return
```

Note that the return statement means the function will stop running code there. You can then setup the file handler outside of the subsequent if's, except for setting the corresponding level.

students/andrew_garcia/lesson02-assignment/charges_calc.py
```diff
@@ -0,0 +1,118 @@
...
+    with open(filename, 'w') as file:
+        json.dump(data, file)
+
+
+if __name__ == "__main__":
+    ARGS = parse_cmd_arguments()
+    debug_setup(ARGS.debug)
+    DATA = load_rentals_file(ARGS.input)
+    DATA = calculate_additional_fields(DATA)
+    save_to_json(ARGS.output, DATA)
```
ldconejo, line 118: The function could be simplified to something like this:

```
def debug_setup(debug_level):

    if debug_level == 0:
        logger.disabled = True
        return

    """ Sets the logging file name and levels """
    # create log format to use for logging
    log_format = '%(asctime)s %(filename)s:%(lineno)-3d %(levelname)s %(message)s'
    formatter = logging.Formatter(log_format)

    # create and set file name for log file
    log_file = datetime.datetime.now().strftime('%Y-%m-%d')+'.log'

    console_handler = logging.StreamHandler()
    console_handler.setFormatter(formatter)

    logger = logging.getLogger()
    logger.addHandler(console_handler)

    file_handler = logging.FileHandler(log_file)
    file_handler.setFormatter(formatter)

    if debug_level == 1:
        logger.setLevel(logging.ERROR)
        file_handler.setLevel(logging.ERROR)
        console_handler.setLevel(logging.ERROR)
        logger.addHandler(file_handler)
        return

    if debug_level == 2:
        logger.setLevel(logging.WARNING)
        file_handler.setLevel(logging.WARNING)
        console_handler.setLevel(logging.WARNING)
        logger.addHandler(file_handler)
        return

    if debug_level == 3:
        logger.setLevel(logging.DEBUG)
        file_handler.setLevel(logging.DEBUG)
        console_handler.setLevel(logging.DEBUG)
        logger.addHandler(file_handler)
        return

    logger.error('Incorrect Debug Level')
    raise ValueError
```

PR #321, opened 2019-11-17
lesson07 final

franjaku:


franjaku:
Now includes lesson 08 files as well.

PR #322, opened 2019-11-18
Lesson01 assignment commit

amirgnyawali:


ldconejo:
Great first assignment! Pylint graded your code at 10/10, good use of Pylint rule disabling, coverage is at least 90% for all files.

PR #328, opened 2019-11-20
lesson09

franjaku:


ldconejo:
Hi Frank,

For the decorator exercise, there a couple of things to correct:

1. Your _logging_decorator()_ function is currently not making any decisions based on a command line parameter, instead, it disables logging right away. Try to change it to use a parameter to maybe force a decorated function into DEBUG mode for logging purposes.

2. You're not decorating any functions, which is the idea of the exercise. Once #1 is implemented, I would expect to see something like this on your code:

```
@logging_decorator
def calculate_additional_fields(data):
.
.
.
```

ldconejo:
The other two implementations are good.

Also, make sure all three parts are getting 10 out of 10 on Pylint, sparsely use Pylint disable statements for those Pylint warnings which do no make sense for the purposes of your code.

Regards,

Luis

PR #335, opened 2019-11-24
Initial L01 work

pbehrend:
@ldconejo 
Hi Luis, here's my partially complete L01 assignment. Committing now since I have a question:
When we're creating an object that will be used for testing, for example Inventory(125, 'Test Item', 50, 80), should it be assigned to a simple variable name in each test method definition or somehow defined as self in __init__?
This doesn't seem to work since it overrides the TestCase init:
def __init__(self): 
    self.inv = Inventory(125, 'Test Item', 50, 80)

### DIFF
--- a/students/philip_behrend/lesson01/test_unit.py
+++ b/students/philip_behrend/lesson01/test_unit.py
@@ -1,0 +1,27 @@
+from inventory_management.product_classes import Inventory, Furniture, ElectricAppliances
+import inventory_management.main
+from unittest import TestCase 
+
+class InventoryTests(TestCase):
+    """Test for the Inventory class"""
+
+    def test_inv_init(self):
+        inv = Inventory(125, 'Test Item', 50, 80)
+        self.assertEqual(inv.product_code, 125)
+        self.assertEqual(inv.description, 'Test Item')
+        self.assertEqual(inv.market_price, 50)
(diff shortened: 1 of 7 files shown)


students/philip_behrend/lesson01/test_unit.py
```diff
@@ -0,0 +1,27 @@
+from inventory_management.product_classes import Inventory, Furniture, ElectricAppliances
+import inventory_management.main
+from unittest import TestCase 
+
+class InventoryTests(TestCase):
+    """Test for the Inventory class"""
+
```
ldconejo, line 7: Use _setUp_ to create your instance of _Inventory_, which you can then reuse across _InventoryTests_. Note that it's being created as an instance variable of _InventoryTests_, hence the use of _self.inv_ instead of just _inv_, that way it can be accessed by each test.
```
    def setUp(self):
        self.inv = Inventory('125', "Test Item", '50', '80')
```
Each test, in turn, will need to to also refer to it as _self.inv_, for example:

```
    def test_inv_init(self):
        inv = Inventory(125, 'Test Item', 50, 80)
        self.assertEqual(self.inv.product_code, 125)
```

ldconejo:
Hi Philip,

I just added a comment on your code, related to your question. Let me know if that helps. I look forward to reviewing your submission for assignment 1.

Regards,

Luis

PR #340, opened 2019-11-26
Committing lesson2 files

amirgnyawali:
To fix the file, I first set up the logging format with the file handler, console handler, and the logger. I then added the debug argument. I then added error handling to the file opener and added a debug message to see which file was opening. I then added a level setter in the main script to set the debug level and set a debug message for the input, output, and logging file. I then added debug messages to each value that came up in calculate additional fields. I also changed the value loop to key, value to keep track of the rental IDs, which I included in all the debug messages. I then added an error if number of days was a negative number. I then added error logic to account for taking a square root of a negative number. I then added for dividing a number by zero. I then added a debug message for a file that faulted. I then added logic to account for a blank rental end date.


ldconejo:
Good job on this one, Amir. One extra thing to consider, although not required for the assignment, is to code it so that no log file is created until the first entry is added (i.e., no log file if not activity to log into it). 

Regards,

Luis

PR #349, opened 2019-11-30
lesson 06 submission

au-gold:
lesson 06 submission


students/mint_k/lesson06/good_perf.py
```diff
@@ -0,0 +1,52 @@
...
+import csv
+
+
+def analyze(filename):
+    """this is to hash through the data"""
+    start = datetime.datetime.now()
+    with open(filename) as csvfile:
+        reader = csv.reader(csvfile, delimiter=',', quotechar='"')
+        found = 0
+        year_count = {
```
ldconejo, line 16: You could simplify it here by making it figure out which years are present in the data:

```
        year_count = {}
        for row in reader:
            try:
                year_count[row[5][-4:]] += 1
            except KeyError:
                year_count[row[5][-4:]] = 1
                continue
```

PR #352, opened 2019-12-01
Lesson 1 Activity and Assignment

bplanica:
Please let me know if there is anything I can improve on - the instruction said not to modify the market prices module, but the module said to add a forced exception, so I was not sure if the exception should be added or not. 

For the pylint message, 'too many positional arguments', is this something we should be concerned about and if so, is there a way around this? 

Additionally, is the patch method acceptable for testing user input in place of using magicmock or mock directly? I got stuck for a while trying to work with magicmock, but found patch on stack overflow - are there any supplemental reading/videos on using magicmock you would recommend?

### DIFF
--- a/students/bplanica/lesson01/activity/calculator/adder.py
+++ b/students/bplanica/lesson01/activity/calculator/adder.py
@@ -1,0 +1,12 @@
+"""
+This module provides an addition operator
+"""
+
+
+class Adder():
+    """adds two operands together, returns result"""
+
+    @staticmethod
+    def calc(operand_1, operand_2):
+        """performs calculation"""
+        return operand_1 + operand_2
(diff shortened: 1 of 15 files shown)


ldconejo:
Hi Breeanna,

Code looks good. In _market_prices_, the comment about raising an exception is an alternative to force the use of _Mock_ when calling _market_prices.py_, since in a real-world scenario this module would be pulling prices from an online source. Don't worry about it.

Yes, _patch_  is the best option for mocking the _input()_ builtin, it is still part of the Mock library.

For the Pylint warnings, there are two things you can do when you run into warnings that do no apply to your code, such as _Too many arguments_ and _Too few public methods_:

- Use a custom _.pylintrc_ file, disabling those warnings for all files in your project (make sure you include this file as part of your pull request.

- Use Pylint disable statements, which you add as comments to specific files in your project. For example:

`# pylint: disable=R0903, R0913`

Although the latter method requires you to disable the warnings for each file in your code, it has the advantage of being part of your code (as opposed to being a separate file).

Regards,

Luis

PR #353, opened 2019-12-02
L01 and L02 revisions

pbehrend:
90% coverage achieved for L01
Logging included in L02 logic


ldconejo:
Hi Phillip,

For lesson 1, code looks good. Please re-enable the following Pylint warnings and make the necessary changes to your code to (i.e., add docstrings). They're very simple formal aspects that should not need disabling:

C0111: Missing module docstring (missing-docstring)
C0304: Final newline missing (missing-final-newline)

Regards,

Luis

students/philip_behrend/lesson02/charges_calc.py
```diff
@@ -0,0 +1,86 @@
...
+    log_file = datetime.datetime.now().strftime("%Y-%m-%d")+'.log'
+    options = {
+        0: logging.CRITICAL,
+        1: logging.ERROR,
+        2: logging.WARNING,
+        3: logging.DEBUG
+    }
+
+    try: 
+        log_level = options.get(int(log_level))
```
ldconejo, line 22: This is a good idea, except that using the _get()_ method will return _None_ for invalid keys instead of raising a _KeyError_ exception. Instead, use square brackets:

`log_level = options[int(log_level)]`

There a minor coding error that will show up once you fix the one above. Should be an easy fix.

PR #361, opened 2019-12-06
L02 revision and L03 assignment

pbehrend:
L02: Adhered to pylint standards

L03: Assignment submitted. Disabled broad exception warning since we're printing Exception as e.
Need help with AssertRaises for test_add_invalid_customer - when I run the code, it shows a peewee.IntegrityError but the AssertRaises fails to raise this error.

Thanks,
Philip


students/philip_behrend/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,49 @@
...
+logger = logging.getLogger(__name__)
+
+def add_customer(**kwargs):
+    """ Add customer function """
+    try:
+        with db.transaction():
+            new_cust = (Customer.create)(**kwargs)
+            new_cust.save()
+            logger.info('Customer addition successful')
+    except Exception as e:
```
ldconejo, line 16: Avoid using catch all exceptions. They're ok for debug code (when you simply need things to pass for a very restricted set of test cases), but the pose a risk on the long run, in that new issues not observed during the development process are inadvertently, an incorrectly, handled by the general exception. Replace it instead with a specific exception(s) that you know can occur or remove the try/except block altogether.

students/philip_behrend/lesson03/assignment/test_operations.py
```diff
@@ -0,0 +1,50 @@
...
+          phone_no='509-888-9999',
+          email='john.deere@gmail.com',
+          status=1,
+          credit_limit=4000)
+        cust = Customer.get_by_id(9)
+        self.assertEqual(cust.firstname, 'John')
+        self.assertEqual(cust.phone_no, '509-888-9999')
+
+    def test_add_invalid_customer(self):
+        with self.assertRaises(IntegrityError) as cm:
```
ldconejo, line 25: This test not working is a side effect of the general exception handling in _basic_operations.py_. The exception is not being raised (which is what the test is expecting) because you _Except_ block catches all exceptions (including _IntegrityError_).

ldconejo:
Hi Philip,

I added a couple of comments on your code, regarding the issues you're having. Additional to that, make sure your _test_operations.py_ is also able to create / delete the test database before and after testing, respectively.

Regards,

Luis

students/philip_behrend/lesson03/assignment/test_operations.py
pbehrend: Hi Luis, thanks for the pointer. I modified the try-except block to catch the IntegrityError specifically (given your statement above). However, the traceback states that there was a _separate_ exception in _new_cust = (Customer.create)(**kwargs)_. 
Do you have advice on how to troubleshoot why the try-except block isn't catching the seemingly obvious single exception (lastname = Null)?


def add_customer(**kwargs):
    """ Add customer function """
    try:
        with db.transaction():
            **new_cust = (Customer.create)(**kwargs)**
            new_cust.save()
            logger.info('Customer addition successful')
    except IntegrityError as e:
        logger.info('Error adding customer')
        logger.info(e)
        logger.info(traceback.format_exc())

PR #368, opened 2019-12-08
Adding in lesson07

cholso:
Finished modules linear.py and parallel.py and written down my findings in notes.txt.


students/chelsea_nayan/lesson07/parallel.py
```diff
@@ -0,0 +1,244 @@
...
+    threads.append(threading.Thread(target=import_products,
+                                    args=(path, 'products.csv'), daemon=True))
+    threads.append(threading.Thread(target=import_customers,
+                                    args=(path, 'customers.csv'), daemon=True))
+    threads.append(threading.Thread(target=import_rentals,
+                                    args=(path, 'rentals.csv'), daemon=True))
+
+    for thread in threads:
+        thread.start()
+        queue_storage.append(output_queue.get())
```
ldconejo, line 235: You don't need _queue_storage_. Queues are thread-safe, which means that different threads can execute _put_ operations on the same queue without trampling each other. You might want to add an additional field to the tuple you're putting into the queue, in order to identify it (for example "Product results", "Customer results", etc.), since threads can finish in any order.

ldconejo: This something like this and compare performance:

```
    for thread in threads:
        thread.start()
        #queue_storage.append(output_queue.get())
    for thread in threads:
        thread.join()
```

You need to launch a separate _join_ in a different for loop to make sure all threads finish. It cannot be in the same operation where the threads are launched, or you will away force it into a linear operation.

Not to spoil it, but these are the timing results I got:

Linear: 3.036235809326172
Parallel: 1.639624834060669

ldconejo:
Hi Chelsea,

See my notes on your code on some things that need to be changed. The way it is currently implemented, your threads are still running serially.

Regards,

Luis

PR #371, opened 2019-12-09
completed lesson 3 assignment, ready for review

TrkCat:
Lesson 3 assignment for review. I realize the unit tests are not true unit tests as they have a connection to a test database... this seemed necessary to me to fully validate the functions, but if there is a better way to do this (just check calls to the peewee functions for example) please let me know.

Thanks, Joel


ldconejo:
Hi Joel,

You make a fair point about having to connect to the database. I would still see this as a valid approach, the creation of a test database. There are way to mock the database, but in my opinion not worth the effort if you can simply use a test setup.

-Luis

ldconejo:
Remember to also submit on the course website (and send me a reminder via email or comment on this pull request to let me know).

TrkCat:
Thanks much Luis. Pull request link added for assignment 3.

Thanks,
Joel

PR #374, opened 2019-12-10
Committing lesson05 files

amirgnyawali:


students/amirg/lesson05/assignment/test_database.py
```diff
@@ -0,0 +1,69 @@
...
+
+class DatabaseTests(TestCase):
+    """
+    Tests for the database module
+    """
+    def test_import_data(self):
+        """
+        Tests the import_data function
+        """
+        directory_path = r"C:/Users/Amir G/SP_Python220B_2019/students/amirg/" \
```
ldconejo, line 17: Try use instead relative paths. Using hard-coded paths like in this case forces anyone trying to run your tests to hunt for these paths to replace them with the appropriate locations (that was my case).

ldconejo:
Hi Amir,

Make sure you setup your _test_database.py_ to clear all existing data in the MongoDB collections you're using, so that your tests are running on a clean test database. Right now, _test_database.py_ is not passing. 

Also, see my comment on using hard-coded paths in multiple locations. Please submit a new pull request once the required additions have been made. Let me know if you have any questions.

Regards,

Luis

PR #382, opened 2019-12-14
Assignment 9 help

gnar5tar:
Part 1 & 3 for assignment 9 are complete. For part 3 I started with os.walk which seems to complete the assignment without recursion so its still in there but I made a new function using os.listdir that uses recursion to complete the assignment. 

I could use a little direction on part 2 of the assignment though. It seems like I have already used a context manager for each function of the database.py. What else should I add?

Thanks!


ldconejo:
Hi Zach,

If you were already using context managers in your database implementation, then you're done with that portion. On the directory walkthrough, the idea is to do it by using recursion, which you're doing in the second function included in the discovery script, so you're good on that account as well.

-Luis

gnar5tar:
Great, thank you!

PR #395, opened 2019-12-16
Lesson 5 Assignment

TrkCat:
Alright, I think I've got it this time. Issue appears to have arisen from the fact that mocking the DATABASE value (database name) was only affecting parts of the program, based on the fact the default function arguments on others were being evaluated at import. Corrected by passing an argument to override the defaults during the test.

As mentioned, this was running successfully on my machine based on the fact the show_available_products and show_rentals functions were actually calling an incorrect database during the test that had already been populated previously.


ldconejo:
Good job on this one, everything worked this time!

-Luis

PR #405, opened 2019-12-18
L06: Pylint and Unittest Revisions

pbehrend:
Hi Luis, a couple of general questions:
1. How are you running coverage to check our assignments? If I run the following command, the coverage values are incorrect:
   coverage run test_perf.py 
   coverage report

2. In general, should we develop tests for our helper files like expand_data.py? 

Thanks,
Philip

PS: No tests developed for poor_perf.py since it was a given function


ldconejo:
Hi Philip,

For this assignment, you don't really need to create unit tests, good and poor performance scripts are evaluated by running them. In general, I run coverage with this command:

`python -m coverage run --include=<file that you want test coverage for> -m unittest < unit test file>`

Another choice is:

`python3 -m coverage run --source=<folder with code you want to measure coverage for> -m unittest <unit test file>`

The second command line will check coverage for all Python scripts within a folder.

-Luis

PR #407, opened 2019-12-19
Assignment 2 - adding logging to existing module

bplanica:
Please let me know if anything could be improved.

Thanks! Breeanna

### DIFF
--- a/students/bplanica/lesson02/assignment/code/charges_calc.py
+++ b/students/bplanica/lesson02/assignment/code/charges_calc.py
@@ -1,0 +1,114 @@
+"""
+Returns total price paid for individual rentals
+12/18/2019, BPA, added logging/debuging
+"""
+
+import argparse
+import json
+import datetime
+import math
+
+import sys
+import logging
+LOG_FORMAT = "%(asctime)s %(filename)s:%(lineno)-3d %(levelname)s %(message)s"
+LOG_FILE = datetime.datetime.now().strftime("%Y-%m-%d") + '.log'
+
+FORMATTER = logging.Formatter(LOG_FORMAT)
+
+FILE_HANDLER = logging.FileHandler(LOG_FILE)
+FILE_HANDLER.setLevel(logging.WARNING)
+FILE_HANDLER.setFormatter(FORMATTER)
+
+CONSOLE_HANDLER = logging.StreamHandler()
+CONSOLE_HANDLER.setLevel(logging.DEBUG)
+CONSOLE_HANDLER.setFormatter(FORMATTER)
+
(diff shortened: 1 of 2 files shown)


ldconejo:
Hi Breeanna,

Code looks good, one thing to correct is the use of exceptions, you are using "catch all" exceptions in a couple of places in your code, meaning that _any_ exception happening on those places, regardless of the type, will be handled within the _except_ block. You are using _sys.exit()_, which will prevent your code from running any further and call attention to the issue.

However, it's still a risk to be handling all exceptions with the same code, as once your code is in production, it might run into a different type of issue (which would be handled by the _except_ block and potentially be ignored.

I'd recommend switching to specific exception types, based on known potential issues (I will add an example directly in your code).

Also, remember to run Pylint on your code an address any PEP8 compliance issues. Submit a new pull request once these changes are implemented.

Regards,

Luis

students/bplanica/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,114 @@
...
+def parse_cmd_arguments():
+    """parse command string arguments"""
+    parser = argparse.ArgumentParser(description='Process some integers.')
+    parser.add_argument('-i', '--input', help='input JSON file', required=True)
+    parser.add_argument('-o', '--output', help='ouput JSON file', required=True)
+    parser.add_argument('-d', '--debug', help='debug enabled/level', default=0)
+    return parser.parse_args()
+
+
+def load_rentals_file(filename):
```
ldconejo, line 42: This could be changed to this, addressing one specific exception type that's quite likely to occur:

```
def load_rentals_file(filename):
    """import data"""
    with open(filename) as file:
        try:
            data = json.load(file)
            logging.info("Data has been sucessfully imported")
        except json.decoder.JSONDecodeError as exc:
            logging.error(f"JSON source file contains formatting errors; Exception: {exc}")
            sys.exit()
    return data
```

students/bplanica/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,114 @@
...
+                    logging.error(f"{key}: invalid total price ({value['total_price']})" +
+                                  f", squareroot returns imaginary value")
+
+            logging.debug("Calculate cost per unit")
+            try:
+                value['unit_cost'] = value['total_price'] / value['units_rented']
+            except ZeroDivisionError:
+                logging.error(f"{key}: invalid units rented ({value['units_rented']})")
+
+        except Exception as exc:
```
ldconejo, line 95: This general exception should also be replaced.

PR #411, opened 2019-12-22
L08 and L09 Assignments

pbehrend:


ldconejo:
Hi Philip,

For the script to find PNG files in lesson 9, the idea is that you implement it using recursion. The solution you use works, but it abstracts that portion. Could you give it another try using recursion?

Thanks,

Luis

PR #415, opened 2019-12-24
L07 with updated threading & L09 with branching issue

pbehrend:
Hi Luis, I'm facing issues in L09 with the recursion logic. Even though I'm setting the absolute path for the subdirectories, the IF statement for os.path.isdir(os.path.abspath(item)) isn't recognizing these subdirectories as folders (Relevant lines: 14, 22 in pngdiscover.py)

### DIFF
--- a/students/philip_behrend/lesson07/assignment/parallel.py
+++ b/students/philip_behrend/lesson07/assignment/parallel.py
@@ -1,11 +1,11 @@
 """ Database file for HP Norton furniture store """
 
 import logging
-import pandas as pd
-from pymongo import MongoClient
 from datetime import datetime
+from multiprocessing import Queue
 from threading import Thread
-from multiprocessing import Queue 
+from pymongo import MongoClient
+import pandas as pd
 
 LOGGER = logging.getLogger()
 LOGGER.setLevel(logging.INFO)
@@ -39,11 +39,6 @@ def reset_db():
     """ This module deletes all records in the database"""
     mongo = MongoDBConnection()
 
-    start_prod = 0
-    end_prod = 0
-    start_cust = 0
-    end_cust = 0
-
     with mongo:
         # Mongo database connection
         db = mongo.connection.NortonFurniture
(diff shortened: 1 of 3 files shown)


students/philip_behrend/lesson07/assignment/parallel.py
```diff
@@ -143,23 +140,22 @@ def show_rentals(product_id):
...
-    prod_thread.start()
-    cust_thread = Thread(target=import_file, args=(out_data, 'customers.csv',1))
-    cust_thread.start()
-    rent_thread = Thread(target=import_file, args=(out_data, 'rentals.csv',2))
-    rent_thread.start()
-
-    prod_thread.join()
-    cust_thread.join()
-    rent_thread.join()
-    prod_output = out_data.get()
```
ldconejo, line 158: Hi Philip,

There's one issue with popping results out of the _out_data_ queue in this way. You're assuming the results will always go into the queue in the same order, which is not guaranteed. Whichever thread finishes first will be the one pushing results into the queue and that could be any of the three threads. Think of a different way of figuring out which thread the result popping out of the queue comes from. 

Also, you can use _while not queue.empty():_ to iterate through the queue, although doing it with three code lines as you're doing it here is also fine.

-Luis

students/philip_behrend/lesson09/assignment/pngdiscover.py
```diff
@@ -9,8 +9,34 @@ def parse_args():
     parser.add_argument('-i', '--input', help="Input directory", required=True)
     return parser.parse_args()
 
-def find_png(parent_dir, png_list):
+def find_png(parent_dir, output_list):
     """ Find all PNG files within parent directory and return list of file names """
+    os.chdir(parent_dir)
```
ldconejo, line 14: You don't need to change directories within your script, just provide the right path for _os.listdir_ to work with.

students/philip_behrend/lesson09/assignment/pngdiscover.py
```diff
@@ -9,8 +9,34 @@ def parse_args():
...
+    png_list = []
+    for item in os.listdir(parent_dir):
+        print(f"ITEM: {item}")
+        print(os.listdir(parent_dir))
+        if item.endswith('.png'):
+            print(f'PNG:{item}')
+            png_list.append(item)
+        elif os.path.isdir(os.path.abspath(item)):
```
ldconejo, line 22: Try _os.path.join_ instead:

`elif os.path.isdir(os.path.join(parent_dir, item)):`

students/philip_behrend/lesson09/assignment/pngdiscover.py
```diff
@@ -9,8 +9,34 @@ def parse_args():
...
+            subdir = os.path.abspath(item)
```
ldconejo, line 23: Same here:

`subdir = os.path.join(parent_dir, item)`

ldconejo:
Hi Philip,

I added some comments to both lesson 7 and lesson 9. For lesson 9, those changes got it working for me, you were on the right path (figuratively and almost literally :-) ).

Happy holidays!

-Luis

PR #417, opened 2019-12-26
submission for assignment #1

humbertojrg:


ldconejo:
Hi Humberto,

Your pull request does not include any files to perform unit and integration testing for your code, which is an important part of the assignment. Please make sure to include those in a new pull request. Please let me know if you have any questions.

Best regards,

Luis

humbertojrg:
I have my main files in a folder called 'inventory_management' and my test files outside of that. I've tried adding them all but it seems like it won't add then commit them all. Any suggestions?

humbertojrg:
> Hi Humberto,
> 
> Your pull request does not include any files to perform unit and integration testing for your code, which is an important part of the assignment. Please make sure to include those in a new pull request. Please let me know if you have any questions.
> 
> Best regards,
> 
> Luis

Hi Luis,

I have my main files in a folder called 'inventory_management' and my test files outside of that. I've tried adding them all but it seems like it won't add then commit them all. Any suggestions?

PR #422, opened 2019-12-27
Submitter for assignment 01

humbertojrg:


ldconejo:
Hi Humberto,

Looks like all files made it this time :-) 

A couple of things: Coverage for _main.py_ is below the 90% target, so will need to add additional tests. You can run _python coverage html_ to get an html report showing which code lines are not covered.

You also need to run Pylint on your code and address any PEP8 compliance issues reported.

Once again, let me know if you have any questions.

Regards,

Luis

PR #424, opened 2019-12-28
lesson05 assignment completed

brianminsk:
To run the tests: "python -m unittest test_database.py"


students/brian_minsk/lesson05/assignment/database.py
```diff
@@ -0,0 +1,177 @@
...
+    errors that occurred, in the same order.
+
+    Keyword arguments:
+    directory_name - the directory that contains the csv_files
+    product_file - a csv file with product data with the first line containing the field names
+    customer_file - a csv file with customer data with the first line containing the field
+                    names
+    rentals_file - a csv file with rental data with the first line containing the field names
+    """
+    n_products, errs_products = populate_collection(directory_name + '\\' + product_file,
```
ldconejo, line 52: The '\\' in these lines of code causes your unit tests to fail on my Mac, because of the different conventions on Unix-like systems, but it was easy for me to fix it for reviewing your assignment.

You might need to consider this in some of your future projects need to work on both Windows and Unix-like operating systems. That said, no need to make any changes for this assignment :-)

brianminsk: Gotcha. I did that quickly and meant to go back and use the Python file path creation functions but I forgot... Thanks for fixing.

PR #425, opened 2019-12-29
Initial Pull Request for Lesson 7

TrkCat:
I've taken a crack at lesson 7, I believe I'm on the right track but would appreciate a review of what is here to determine if this is what the assignment was looking for.

Thanks,
Joel


ldconejo:
Hi Joel,

Looks good, the implementation and justification for using threading are correct. At some point, I think it will be necessary to update this exercise to a case in which the advantages of threading and/or multiprocessing are clearer.

Good luck with lesson 8!

-Luis

P.S.: I'm closing this pull request, instead of merging, so that you can open a new one once you have also entered your submission on the course website.

PR #429, opened 2019-12-30
Customer model, basic operations module and test file for lesson 3

bplanica:
Please let me know if anything can be improved!

I had trouble using except clause to catch the TypeError caused when trying to create a new customer  with a required argument missing - is this something that can be caught in an except clause or will it always be a hard stop when running the code?

### DIFF
--- a/students/bplanica/lesson03/assignment/basic_operations.py
+++ b/students/bplanica/lesson03/assignment/basic_operations.py
@@ -1,0 +1,79 @@
+"""Basic operations utilizing the customer_model"""
+
+import logging
+import peewee as pw
+from customer_model import Customer, DATABASE
+
+logging.basicConfig(level=logging.INFO)
+LOGGER = logging.getLogger(__name__)
+
+# toggle below comment when needed
+DATABASE.drop_tables([Customer])
+DATABASE.create_tables([Customer])
+#DATABASE.close()
+
+def add_customer(name, lastname, home_address, phone_number, email_address,
+                 status, credit_limit):
+    """create a new customer object in the database table"""
+    try:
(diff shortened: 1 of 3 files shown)


ldconejo:
> Please let me know if anything can be improved!
> 
> I had trouble using except clause to catch the TypeError caused when trying to create a new customer with a required argument missing - is this something that can be caught in an except clause or will it always be a hard stop when running the code?

This is actually a great question, you can try this:

```
    def test_a_add_customer_type_error_exception(self):
        with self.assertRaises(TypeError):
            bo.add_customer()
```

Note that I'm encapsulating the call that raises the exception (_bo.add_customer()_) within a context manager, so that it doesn't crash the run itself (it's akin to trapping the exception for unit testing purposes).

All assignment requirements and coverage goal are met, keep up the good work!

-Luis

students/bplanica/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,79 @@
...
+        a_customer.save()
+        LOGGER.info('Database update successful')
+    except pw.DoesNotExist:
+        LOGGER.info('Customer ID provided does not exist')
+
+
+def list_active_customers():
+    """list all active customer: status = 1 (Active), status = 0 (Inactive)"""
+    active_count = 0
+    for _ in Customer.select().where(Customer.status == True):
```
ldconejo, line 75: You don't need to do the actual _Customer.status == True_ comparison, _Customer.status_ will evaluate to true on its own (assuming it has a value of 1), so the line should be:

`    for _ in Customer.select().where(Customer.status):`

PR #430, opened 2019-12-30
submission for assignment 02

humbertojrg:


ldconejo:
Hi Humberto, 

Looks good. One thing that needs to be corrected is that your code accepts invalid logging levels (for example _-d 10_) which should not be the case. Could you take a shot at correcting that?

Thanks,

Luis

PR #431, opened 2019-12-30
Finished assignment 3

ramr24:


ldconejo:
Hi Ramkumar,

The implementation looks good, but your unit testing is failing. Take a look at your _setup_database_ function and see if it's appropriate to run it in every test. 

Additionally, consider using the _setupUp_ and _tearDown_ methods instead of having _setup_database_ outside of the test class. Finally, make sure you run Pylint on your code and address any issues reported.

Regards,

Luis

PR #432, opened 2019-12-31
Added Initial Assignment07 Files

ajaggarcia10:
I've worked on the initial part of having a linear and parallel database, but I'm not sure if I've correctly done the threading portion.

### DIFF
--- a/students/andrew_garcia/lesson07-assignment/parallel.py
+++ b/students/andrew_garcia/lesson07-assignment/parallel.py
@@ -1,0 +1,212 @@
+'''Converting CSV Files and Putting Them Into a Database'''
+
+import csv
+import os
+import logging
+import time
+import threading
+
+
+from pymongo import MongoClient
+
+# pylint: disable=invalid-name
+# pylint: disable=too-many-statements
+# pylint: disable=too-many-locals
+
+logging.basicConfig(level=logging.INFO)
+LOGGER = logging.getLogger(__name__)
+
+
+class MongodbConnection():
+    """ Connects to Mongodb """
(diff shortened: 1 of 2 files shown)


students/andrew_garcia/lesson07-assignment/parallel.py
```diff
@@ -0,0 +1,212 @@
...
+    path = 'D:\\UW\\PY220\\Examples and Assignments\\lesson07\\assignment\\csv_files'
+    threads = []
+
+    threads.append(threading.Thread(target=import_products, args=(path, 'products.csv')))
+    threads.append(threading.Thread(target=import_customers, args=(path, 'customers.csv')))
+    threads.append(threading.Thread(target=import_rentals, args=(path, 'rentals.csv')))
+
+    for thread in threads:
+        thread.start()
+
```
ldconejo, line 209: Hi Andrew,

The implementation of threading looks good, but you're missing one step: Joining.  The three lines of code you have after _thread.start()_ on line 208 will run as soon as the last thread is started and you exit the for loop.

You have conveniently packed all three threads into a single list, _threads_, you can now add another _for_ loop to do the joining:

```
    for thread in threads:
        thread.join()
```

This works as a "meeting point" of sorts for all three threads, to make sure each has finished before (and all data has been processed) before concluding execution. Note that the size of the data set may not be big enough to cause a significant difference between linear and parallel implementations, in particular if you have a relatively new (and fast) system.

Also, remember to run Pylint on _parallel.py_ and fix any outstanding issues.

Let me know if you have any additional questions.

Happy New Year!

-Luis

PR #437, opened 2020-01-02
Resubmit completed assignment 3

ramr24:
Completed and fix the issues with assignment 3. Fixed the test_unit.py and checked through pylint/flake8 and had no issues with any of the files.


ldconejo:
Hi Ramkumar,

Looks good. Coverage is lower than required (82% versus 90%). See if you can add a couple of more tests to get coverage past 90%. If you want to see which specific lines of code are not covered, you can try running:

```
$ python -m coverage run --include=basic_operations.py -m unittest test_unit.py
$ python coverage html
```

Which will create an html report showing code not covered. Also, remember to run Pylint and make sure no outstanding issues are present (you can use _pylint disable_ for specific rules that are not relevant to your code, which could be the case here).

Regards,

Luis

ramr24:
Hi Luis, thank you for the feedback. I will fix the coverage issue and resubmit alongside with submitting assignment 4. So for the pylint issues, I didn't see any major ones besides # pylint: disable=too-few-public-methods on customer_model.py, and # pylint: disable=too-many-arguments
# pylint: disable=logging-format-interpolation on basic_operations.py. Were these still showing up as issues when you checked the assignment? I disabled them but I can change to the correct logging format if needed, I was just using the f string formatting to read it simpler.

PR #446, opened 2020-01-05
Lesson 03 Submission

zmeves:


students/zach_meves/lesson03/basic_operations.py
```diff
@@ -0,0 +1,96 @@
...
+def search_customer(customer_id: int) -> dict:
+    """
+    Return a dictionary representing the customer with the given ID.
+    :param customer_id: int, customer ID
+    :return: dict
+    """
+
+    try:
+        res = _get_by_id(customer_id)
+    except Exception:
```
ldconejo, line 44: Avoid using general exceptions; the risk being that new issues not seen in the development process might be caught by _except Exception_, masking and improperly handling new issues. Instead, replace this block with specific exception types.

students/zach_meves/lesson03/basic_operations.py
```diff
@@ -0,0 +1,96 @@
...
+    Updates an existing customer's credit limit. Raises a ValueError if no such
+    customer exists.
+
+    :param customer_id: int, customer ID
+    :param credit_limit: float, new credit limit
+    """
+
+    try:
+        res = _get_by_id(customer_id)
+    except Exception:
```
ldconejo, line 77: Same case with a general exception.

students/zach_meves/lesson03/basic_operations.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def list_active_customers() -> int:
+    """
+    Return number of customers whose status is currently active.
+
+    :return: int
+    """
+
+    return Customer.select().where(Customer.status == True).count()
```
ldconejo, line 92: Because of the concept of _Truthiness_ in Python, you don't need to have _== True_, it suffices to do:

`return Customer.select().where(Customer.status).count()`

ldconejo:
Hi Zach,

Code looks good and coverage is at 100%, see my comments on your code for some minor changes (also, remember to run Pylint to root out any PEP8 issues).

Regards,

Luis

PR #449, opened 2020-01-05
General exeption removed, replaced with a specific exception.

KRTstudent:


students/kevin_t/Lesson7/parallel.py
```diff
@@ -0,0 +1,141 @@
...
+    THREAD2.start()
+    THREAD3 = threading.Thread(target=import_rentals, args=('input_files', 'rentals_l7.csv',
+                                                            MY_QUEUE))
+    THREAD3.start()
+
+    THREAD1.join()
+    THREAD2.join()
+    THREAD3.join()
+
+    CUSTOMERS = MY_QUEUE.get()
```
ldconejo, line 127: There's a problem with this approach for getting results out of the queue, as not all threads will always return in the _correct_ order, so you might end up getting the _products_ results when trying to get _customers_. You can modify your results so that there's a field or something else that will indicate which thread they came for.

PR #450, opened 2020-01-06
adding in assignment 05

humbertojrg:


students/humberto_gonzalez/lesson05/assignment/database.py
```diff
@@ -0,0 +1,141 @@
...
+            for entry in product_data:
+                try:
+                    product_entry = {"product_id":entry[0],
+                                     "description":entry[1],
+                                     "product_type":entry[2],
+                                     "quantity_available":entry[3]}
+                    products.insert_one(product_entry)
+                    product_count += 1
+                    LOGGER.info(f"Added {entry[0]} to product database")
+                except:
```
ldconejo, line 69: Avoid using bare or general exceptions in your code. They're ok for debugging, when you just need to force your code to run, but they present a significant risk for code intended for release (which is what these assignments try to reproduce), as they will capture _anything_ triggering an exception, including issues not seen during development and which might need different handling.

PR #453, opened 2020-01-06
Lesson 8 homework

KRTstudent:


KRTstudent:
Luis, with this pull request, I also included my HW7 update. It doesn't
look like there was an update to my grade or any comments back. Do I need
to re-push?

PR #459, opened 2020-01-08
Python files for assignment 4

bplanica:
Hello - I could not really find any good place to use an iterator, etc in my assignment 3 file that made more sense than what was already there.

I created a new function to try to add one, which I think works, but I noticed that it did not matter if I had the dunders in the customer model for __inter__ and __next__ -is this alright not to include and/or in what cases would this be required? Will they ever need to be different from the default?

Please let me know if there is anything I can improve on.

### DIFF
--- a/students/bplanica/lesson04/assignment/basic_operations.py
+++ b/students/bplanica/lesson04/assignment/basic_operations.py
@@ -1,0 +1,118 @@
+"""Basic operations utilizing the customer_model"""
+
+import logging
+import peewee as pw
+from customer_model import Customer, DATABASE
+
+# toggle below comment when needed
+DATABASE.drop_tables([Customer])
+DATABASE.create_tables([Customer])
+#DATABASE.close()
+
+def setup_logging():
+    """setup file and console logging"""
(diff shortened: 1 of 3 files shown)


ldconejo:
Hi Bree,

You are already using iterators in _search_all_customers_, which fulfills the requirement of the assignment. Custom iterators are implemented as a class, those require the use of double underscores ("dunders") for the ___iter___ and ___next___ methods. As an example, try this code:

```
class CustomRange:
    def __init__(self, start, end, increment):
        self.current = start
        self.end = end
        self.increment = increment

    def __iter__(self):
        return self

    def __next__(self):
        if self.current < self.end:
            output = self.current
            self.current += self.increment
            return output
        else:
            raise StopIteration()

if __name__ == "__main__":
    test_range = CustomRange(5, 30, 3)
    print(f"First element: {next(test_range)}")
    for element in test_range:
        print(element)
```

Now, try it again, removing the dunders inside of class _CustomRange_.

Back to the assignment, remember you need to keep test coverage at 90%+, it's currently at 75%, try adding some additional tests to improve coverage.

Regards,

Luis

PR #462, opened 2020-01-10
Added assignment 07

humbertojrg:


ldconejo:
Looks good, when creating threads, you can also add them into a list, so that you can later iterate the list once to run _start_ on each thread and then a second time to run _join_.

PR #463, opened 2020-01-10
Assignment 3

nlenssen2013:


ldconejo:
Hi Nick,

This one looks good! Coverage is a little low at 79% (versus the 90%). Could you add test coverage for some or all of the exceptions? That should bring you past 90%.

Thanks,

Luis

PR #467, opened 2020-01-11
completed lesson07

brianminsk:
I suggest reviewing my lesson notes first.


ldconejo:
Hi Brian,

This was easily the most detailed submission I have seen for this assignment so far. Your results show the value that can be obtained from attempting different approaches to optimize existing and how the way modules you're using (such as PyMongo) are architected also influences such results.

Your proposal of attempting multi-processing is also interesting, as it would allow you to skip the GIL and get actual multithreading. Since this is not a CPU-heavy task, the expectation would be that no significant improvements would be observed, but maybe not.

Great work.

-Luis

PR #469, opened 2020-01-12
Anthony McKeever - Lesson 7 Content Submission

Snip3rM00n:
This is my submission for Lesson 7's Activity and Assignment.

I feel like my unit tests the assignment could possibly be done better than how I did it.  Basically the Parallel (class _TestParallel_) and Parallel Contention (class _TestParallelContention_) unit tests are identical except changing which class it calls for testing.  If you have some recommendations on how I could make it more efficient in the future, I'd be happy to hear them. :)

Regardless, coverage for linear.py, parallel.py and parallel_contention.py are at 100%!  😁

**_Squashed Commit History_**

* Initial commit of lesson 7 (activity)

* Complete activity with both locks and semephores

* Assignment - Multithreading HP Norton DB

Create examples of linear processing, parallel processing and parallel
processing with thread locking.
Add customer, product and rental data files
Add Data expander
Add Unit tests

* Implement self feedback


ldconejo:
Hi Anthony,

Looks good. Since you're testing the performance of both _parallel.py_ and _linear.py_, the way you're testing them is very reasonable.

Good job, very detailed submission and good analysis, as usual from you.

Regards,

Luis

PR #471, opened 2020-01-13
Assignment 4 Submission

zmeves:
Test cases refactored to use iterators/generators, primarily in setUp() method, to construct arbitrary number of test customers.


ldconejo:
Very clever to incorporate the new structures as part of unit testing. I think this is the first time a student has come up with such solution, instead of incorporating them into the main code, which is not always straightforward.

-Luis

PR #474, opened 2020-01-13
completed lesson08 assignment

brianminsk:
Minimal code to satisfy assignment requirements. Invoke by typing "python inventory.py" at the command line.

### DIFF
--- a/students/brian_minsk/lesson08/assignment/inventory.py
+++ b/students/brian_minsk/lesson08/assignment/inventory.py
@@ -1,0 +1,47 @@
+""" lesson08 code - Brian Minsk
+"""
+
+import csv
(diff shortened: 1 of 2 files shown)


ldconejo:
Implementation is good. You could've used _partial_ for _single_customer_ but your code works as well.

-Luis

brianminsk:
> Implementation is good. You could've used _partial_ for _single_customer_ but your code works as well.
> 
> -Luis

Doh! I read "_functools.partial_" in the assignment instructions but somehow I forgot to actually use it! Thanks for letting me know.

I've added a new function that uses **both** functools.partial **and** closure and will create a new pull request.

PR #476, opened 2020-01-14
added a function with both functools.partial and a closure

brianminsk:
see the new function called "single_customer_with_partial"


ldconejo:
Thanks for taking the time to put the extra work to give it a try using _partial_, Brian.

-Luis

PR #477, opened 2020-01-15
Adding assignment 09

humbertojrg:


ldconejo:
Hi Humberto,

For _pngdiscovery.py_, the idea is that you implement it using recursion (i.e., a function calling itself in order to explore each folder looking for png files).

Regards,

Luis

PR #486, opened 2020-01-17
Assignment 5

bplanica:
Please let me know if there is anything I can improve on! Thanks, Breeanna


ldconejo:
Looks good, the use of absolute paths gave me some trouble while reviewing, something to consider is that your tests should in general be able to run across different systems.

Other than that, everything is working and coverage is at a very healthy 95%.

-Luis

PR #489, opened 2020-01-18
Fixed Lesson02 errors, completed lesson03

alexanderboone:
Fixed pep8 errors with lesson02 and completed lesson03. Pylint didn't like the wildcard importing (which was used in the lesson videos and in my program) so I had to generate an rc-file to ignore those.


ldconejo:
Hi Alex,

There a few things to correct on lesson 3:

- Coverage of _basic_operations.py_ is low at 71%. If you run an HTML report, you'll see that none of your exceptions are covered, so you should probably add additional tests to trigger exception conditions and test that code. Let me know if you need some guidance on how to test for exceptions.

- On the same topic of exceptions, avoid using general traps (_except Exception as exc_). The problem with this is that everything triggering an exception (including new issues you have not seen during development) will be trapped in those, in some cases hiding new sources of error that might need to be addressed in a different way. Instead, replace these general cases with specific exceptions (or remove the _try/except_ block altogether).

It's ok to use wildcard import for this course purposes, it's not always a good practice, in particular in cases in which the amount of memory you have is limited and you're import every single piece of code from large libraries when you might only need to use a handful of methods.

-Luis

PR #496, opened 2020-01-25
assignment 2

calvinpy:


ldconejo:
Hi Calvin,

Looks good. Only change I would make would be to alert the user when they enter an invalid level for debug (for example, 10). Right now, the program simply runs and no debug log is created, it would be useful for the user to get a warning or some indication that no log will be created. Only a small detail, so no need to resubmit the assignment.

Keep up the good work!

Luis

PR #498, opened 2020-01-25
Lesson 1

theseanhodges:
Edited files and added test_unit.py and test_integration.py


ldconejo:
Very good first assignment, coverage is higher than the 90% target for all files, no issues reported by Pylint. 

-Luis

PR #502, opened 2020-01-27
Finished Assignment 2

ramr24:


ldconejo:
Hi Kumar,

Looks good. Two quick changes to make:

1. Make sure your code will error out if an invalid log level is entered (highest level should be 3).
2. Run Pylint, some minor trailing-whitespace warnings should be fixed.

Thanks,

Luis

PR #505, opened 2020-01-30
Resubmitting Assignment 2

ramr24:
Fixed pylint errors and added an else statement for log level issue.


ldconejo:
Hi Kumar,

Almost correct. For PEP8 compliance, we're good now. For invalid debug levels, you code currently fails for everything except debug level 0. See if you can change so that levels 0, 1, 2 and 3 can be used.

Regards,

Luis

PR #513, opened 2020-02-01
assignment_3 files

calvinpy:


ldconejo:
Hi Calvin,

Implementation looks good. One thing to improve is test coverage. It's currently at 76%, could you add some additional tests to bring it up to 90%+?

Thanks,

Luis

PR #516, opened 2020-02-03
lesson01 assignment

joli-u:
completed hp norton project assignment


ldconejo:
Great first assignment, coverage is higher than 90% for all files and no PEP8 issues. Good job.

-Luis

PR #517, opened 2020-02-03
Add lesson01 test files.

smckellips:


ldconejo:
Hi Sean,

Much better! No major PEP8 issues, only thing to fix is test coverage. _main.py_ is currently at 86%. The goal is for all files to get to 90% or higher. You can run the following to get an HTML coverage report:

$ python -m coverage html

Run it after you complete a coverage analysis, it will create a separate folder that can be viewed in a browser with details on what code lines are not being covered. For _print_ statements, don't bother with capturing standard output, just making sure the code runs should be enough.

-Luis

smckellips:
I did run it, but got odd results on the main file.  It says I skipped the
'else:' line but ran the contents of the else clause.  It also says I
skipped one of the middle lines printing the main menu.  Some of the other
lines it says I skipped were the empty line breaks between functions.  So I
gave up on it.

PR #526, opened 2020-02-07
Lesson04 submisssion

mattcasari:
Not many changes, but adding to logging, added new function and more use of comprehensions & iterables


ldconejo:
Hi Matt,

Looks good, just one thing to correct: Could you replace your general exceptions for specific ones? The issue with _catch all_ exceptions is that they will handle the issues you know about and the ones that could show up later (in which case, the handling might not be the right one). 

Replace them except with specific exception traps; you can also remove the trap in cases in which you don't know which specific exception type you're expecting.

Regards,

Luis

PR #527, opened 2020-02-07
Lesson 04

theseanhodges:


students/shodges/lesson04/basic_operations.py
```diff
@@ -0,0 +1,136 @@
...
+    except (IndexError, DoesNotExist):
+        logging.error('Unable to update record for customer_id == %s', customer_id)
+        return False
+
+
+def list_active_customers():
+    """
+    This function will return the count of active customers.
+    """
+    return Customer.select().where(Customer.is_active == True).count()
```
ldconejo, line 91: You're disabling _singleton-comparison_ warning on Pylint, but the fix it recommends is actually very simple:

`return Customer.select().where(Customer.is_active).count()`

You don't need to explicitly check for _is_active_ being True.

PR #529, opened 2020-02-08
Lesson01 assignment

HawkeyeUW:
--Calculator plus unit and integration testing files


ldconejo:
Good first assignment. Coverage for _main.py_ is right on the edge, 89% versus a target of 90%, but that's ok. Keep up the good work.

-Luis

PR #533, opened 2020-02-10
Assignment 7

bplanica:
For my findings I did not get much of a notable difference between threading and running the code linear, so I'm not sure if I am missing something. Please let me know if there was something I am missing or if there is anything I can improve on!

Thanks,

Breeanna


students/bplanica/lesson07/assignment/parallel.py
```diff
@@ -0,0 +1,210 @@
...
+    QUEUE = Queue()
+    RESULTS = []
+    THREADS = []
+    THREADS.append(threading.Thread(target=import_product, args=("./", "products.csv", QUEUE)))
+    THREADS.append(threading.Thread(target=import_customer, args=("./", "customers.csv", QUEUE)))
+    THREADS.append(threading.Thread(target=import_rental, args=("./", "rentals.csv", QUEUE)))
+
+    for thread in THREADS:
+        thread.start()
+        RESULTS.append(QUEUE.get())
```
ldconejo, line 206: Hi Bree, 

There's a problem with this approach: You're not waiting for the thread to finish and instead are assuming it has already finished by the time you call the _append_ operation. Instead, launch all your threads (the queue is thread-safe, so they'll be able to share it), then run another _for_ loop to _join_ each thread (this will ensure that all your threads have finished). Note that this means that when you run a _get_ operation on your queue, you will not to figure out which thread placed which data in the queue (you might add a string in the data each thread pushes or something else).

Let me know if this makes sense.

Regards,

Luis

ldconejo:
See my comments on one important change to be made.

PR #534, opened 2020-02-10
Assignment 4

calvinpy:


ldconejo:
Hi Calvin,

Looks good. One thing to consider for future assignments (it's ok for this one) is not to use bare exceptions (_except Exception_) and instead have specific exceptions based on observed issues that have been detected during development. This prevents new issues from being captured and handled by exception blocks that were not intended to handle them.

Keep up the good work,

Luis

ldconejo:
Also, I see you haven't submitted your assignment on the course website. Please send me an email once you've do so.

PR #535, opened 2020-02-10
Assignment07

alheadstrong:
Created linear.py from existing database.py, then parallel.py from that.
Both run from a timeit call if name == main and return a list of tuples. Testing for both is in tests.py andva description of timing and parallelization attempts are included in findings.

### DIFF
--- a/students/al_headstrong/lesson07/assignment/linear.py
+++ b/students/al_headstrong/lesson07/assignment/linear.py
@@ -1,0 +1,150 @@
+"""
+Module to establish mongo database, add and view data.
+"""
+# pylint: disable=invalid-name, too-many-locals
+
+import timeit
+import time
+import logging
+import os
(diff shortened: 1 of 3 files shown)


ldconejo:
Hi Alex, 

A couple of things before grading this one: Could you add your notes on the results obtained under _findings.txt_? The file is currently empty. Also, in _parallel.py_, make sure you _join_ the processes before pulling data from the queue, as the join is the guarantee that the process has finished.

Regards,

Luis

alheadstrong:
Thanks for letting me fix those things. I moved the process join and hit
'save' on the findings.txt file, but now I realize I had committed work for
lesson 8 (not complete) that got pushed along with the two changes. Should
I try to untangle the commits, or just finish the work for lesson 8 and
push again? Sorry for the mess!

PR #539, opened 2020-02-10
Submitting lessons for grading

nullconfig:
- Updated lesson 1 based on previous notes for upgrading
- Submitting lesson 2 for grading


ldconejo:
Hi Billy,

The changes to lesson 1 look good. For lesson 2, everything seems to be ok, except that the _debug_ parameter currently doesn't take any values, for example:

```
$ python charges_calc.py -i source.json -o test.json --debug 3
usage: charges_calc.py [-h] -i INPUT -o OUTPUT [-d]
charges_calc.py: error: unrecognized arguments: 3
```

Could you take a look at that and make sure valid values for debug (0, 1, 2, 3) are accepted while any other value is rejected?

Thanks,

Luis

nullconfig:
I was testing it with -ddd to increase the count. :/  I'll work on the recommendations.

PR #540, opened 2020-02-11
Lesson02 Submssion

HawkeyeUW:
Added Log files, debug log and charges_calc files in github


ldconejo:
Hi Jason,

Looks good. Only one thing to correct: Make sure the _--debug_ parameter only takes valid values (i.e., that it doesn't accept a debug level of 10).

Keep up the good work,

Luis

PR #543, opened 2020-02-11
changed the debug behavior to match assignment

nullconfig:


ldconejo:
Great, thanks for taking the time to fix this!

-Luis

PR #551, opened 2020-02-16
Lesson 09

alheadstrong:
Added to mongo context manager in database, used decorators in charges_calc
and used recursion for png search.


ldconejo:
Nice touch of using an exception to skip trying to open files as if they were directories :-)

PR #555, opened 2020-02-17
lesson02 assignment

stevetwalker:


ldconejo:
Good job! It's a nice touch that the debug level is added to the log file name.

-Luis

stevetwalker:
Hi,

I'm ready to complete assignment four, but I'm struggling to see
opportunities to improve the code using comprehensions, iterators or
generators without adding additional functionality. I could force those
tools into some functions (like the list_active_customers function), but I
feel doing so would be less efficient than using the built-in count method.

Is there anything I'm missing? As an alternative, would it be ok if I write
my own program using these tools and uploaded that instead? On the side I'm
writing a program that generates an ~infinite number of math problems for
any inputted equation that meet certain criteria (e.g. returning all
Pythagorean theorem problems that only use integers with side / hypotenuse
values between 1 and 200). It has plenty of opportunities to incorporate
tools like this.

-Steve

ldconejo:
Hi Steve,

I agree that requirement can feel a bit forced into the existing code. Either way you're proposing is fine: You an force it into an existing function or create new code to use it.

Regards,

Luis

stevetwalker:
Thanks, Luis. That's helpful! In that case, is adding a function like this
enough? (see image below for color formatted version)

  def list_credit_limits():
    """Return a list of active customers along with their credit limits."""

    try:
        connect_customer_db()

        # Creates a list of all active customers
        active_customers = Customer.select().where(Customer.active_customer)

        # Lists credit limits for those customers
        logger.info(f"Listing credit limits for all active customers.")
        return [f"{cust.last_name}, {cust.first_name}: {cust.credit_limit}"
                for cust in active_customers]

    finally:
        db.close()

[image: image.png]

PR #565, opened 2020-02-19
Assignment 3 submission

nullconfig:


students/billy_galloway/lesson_3/assignment/customer_db/basic_operations.py
```diff
@@ -0,0 +1,90 @@
...
+                name=customer[NAME],
+                last_name=customer[LAST_NAME],
+                home_address=customer[HOME_ADDRESS],
+                email_address=customer[EMAIL_ADDRESS],
+                phone_number=customer[PHONE],
+                status=customer[STATUS],
+                credit_limit=customer[CREDIT_LIMIT])
+            new_customer.save()
+            logger.info(f'{name} {last_name} added to database')
+    except Exception as e:
```
ldconejo, line 43: Avoid using bare exceptions like this. A general exception trap will capture (and handled) anything that raises an exception, which poses a significant risk for released code. No need to make a change for this assignment, but keep it in mind for future submissions.

-Luis

PR #572, opened 2020-02-22
Lesson 07 submission

mattcasari:
All files for lesson07


ldconejo:
Hi Matt,

I would give this one another try. With the amount of time left, focus on assignment 8, 9 and 10, then come back to this one.

You have one function in _linear.py_ that is heavy on I/O operations, _import_data()_. I would suggest using threading, so that importing the data from _product_file_, _customer_file_ and _rentals_file_ is not done linearly. Consider the use of queues (which are thread safe, so that you don't need to worry about semaphores) for each import operation to report its results.

Regards,

Luis

PR #580, opened 2020-02-24
Lesson 9 files

bplanica:
Please let me know if there is anything I can improve! 

I couldn't think of much other than a counter for the context manager - out of curiosity, what other things can be done in this case/added that would be useful? 

Thanks! 

Breeanna


ldconejo:
Hi Breeanna,

Try to use the database context manager more extensively. One example would be the _import_data()_ function, but in general any function making use of the database should be converted to use the context manager.

Regards,

Luis

PR #581, opened 2020-02-25
Lesson 9

theseanhodges:


ldconejo:
Thanks for adding details on why you were disabling the _too-many-function-args_ warning. Also, you implementation of the decorator was very nice, only executing logger setup when logging is enabled.

-Luis

PR #588, opened 2020-02-29
Completed lesson09 assignment

alexanderboone:
Completed all three requirements for the lesson09 assignment. Let me know if any improvements can be made.


ldconejo:
Hi Alex,

Your implementation for discovery of png files in a directory tree works; however, the purpose of the assignment is to complete the task using recursion. Could you change it accordingly? The other two parts of the assignment look good.

Regards,

Luis

PR #596, opened 2020-03-05
lesson7

calvinpy:


ldconejo:
Hi Calvin,

It's good that you implemented different methodologies. Two things to correct:

- For threading and multiprocessing, you need to _join_ the threads and the processes you launch, otherwise your code will continue without waiting for all threads to finish. Could you fix that?
- Please add your notes on the performance differences observed compared to linear.

Regards,

Luis

PR #607, opened 2020-03-08
Lesson03 Assignment

stevetwalker:


stevetwalker:
Added list_credit_limits function to basic_operations.py for lesson04.

ldconejo:
Hi Steve,

Looks good, test coverage is lower than the required 90%. Could you add additional tests to get past 90%? 

Regards,

Luis

stevetwalker:
Hi Luis,

Thanks for that reminder. I've updated my coverage so it exceeds 90%. I'm
not sure why I'm not getting 100% coverage, as I have a test in place for
the remaining missing lines of code. Any suggestions?

-Steve

PR #608, opened 2020-03-08
Lesson09 Final Assignment

HawkeyeUW:
Submitted all three assignments + activity


ldconejo:
Hi Jason,

Implementation of decorator and context manager is good. For the PNG discovery program, _os.walk()_ does work, but the idea of the assignment is to do the file discovery using recursion. Could you change the code accordingly?

Thanks,

Luis

PR #609, opened 2020-03-08
resubmission of assignment2

cjfortu:


ldconejo:
Hi Clemente,

This pull request still deletes _charges_calc.py_ and it _source.json_. In your local fork, could you restore both files, commit and then start a new pull request? That should take care of the deleted files.

Thanks,

Luis

ldconejo:
_out.json_ needs to be restored as well.

PR #610, opened 2020-03-08
Lesson05

smckellips:
Lost commit history when I got local repo in an inconsistent state and had to do a hard reset.


ldconejo:
Hi Sean,

Code looks good, with test coverage at 95%. Could you submit on the course website so that I can grade it? Please send me an email one you have done so.

Thanks,

Luis

smckellips:
I had clicked save, but not submit :/  Submitted now.

PR #619, opened 2020-03-11
Files for assignment 9 are completed

nlenssen2013:


ldconejo:
Hi Nick,

Looks good for the decorator and context manager part. For _pngdiscover.py_, you need to implement the file search using recursion. _os.walk()_ is a valid solution to the problem, but not to the goal of solving it with recursion. Could you take another stab at it?

Thansk,

Luis

PR #621, opened 2020-03-12
Lesson 1

Eric74nyc:
Luis,
Thanks for the import guidance, I was able to get the main program to run. However, still running into an issue...I get import errors in VSCode, and pylint returns import-errors (6.52/10). However, I can run main.py from VSCode terminal, and by my machine terminal (using python3.5). So, I cannot figure out why import errors are returned, when the program works, and all unittests pass.

Lastly, I was able to use coverage on my python3.5 installation, and received 94% total. I cannot get it to work on python3.7...I would have to uninstall/reinstall 3.7. Will likely just stay with python3.5 for now.

Thanks for the help!

Eric


ldconejo:
Hi Eric,

I'm glad you got it working. I've generally had inconsistent experiences with IDEs. I do use VSCode, but any code execution I prefer to do it manually in the integrated terminal. Keep up the good work and good luck with assignment 2!

-Luis

PR #629, opened 2020-03-15
Merging lesson_01 with unit tests and pylint assignment.

ghost:
Assignment 1 for pylint and coverage of Inventory Management.


ldconejo:
Good first assignment, test coverage is higher than 90% for all files and no PEP8 compliance issues reported.

PR #632, opened 2020-03-19
Lesson 7 submission

nullconfig:


students/billy_galloway/lesson_7/assignment/hp_norton_inventory/parallel.py
```diff
@@ -0,0 +1,142 @@
...
+            db_collections = hpnorton_db[name]
+            try:
+                # verify that the file exists first
+                if os.path.isfile(f'{directory_name}/{file_names[i]}'):
+                    generate_thread = threading.Thread(target=csv_handler.generate_document_list,
+                                                       args=[f'{directory_name}/{file_names[i]}', name, db_queue])
+                    # start thread and send the
+                    # return value to the queue
+                    generate_thread.start()
+                    # generate_thread.join()
```
ldconejo, line 58: You do need to do a join for each of the threads you launch. Otherwise, code execution will simply continue, without waiting for running threads to finish. What you can do is group all your threads into a list:

thread_list = []
# For each thread, after they have started, do this:
thread_list.append(generate_thread)

Make sure you join all started threads before collecting your results with calls to _db_queue.get()_:

for thread in thread_list:
    thread.join()

Note that this means you don't know which thread will push its results into the queue first, so you'll need to figure out a way to mark the results so that you can identify where each set of results came from.

ldconejo:
Hi Billy,

See my comments on your code for one important change.

Regards,

Luis

PR #634, opened 2020-03-20
test file with updates

FreemaGw:


ldconejo:
Hi Ben,

I see the issue now. I normally don't require students to lint their tests, just the main code. That said, it's good that you are making sure the your tests are also PEP8-compliant. Here's a quick fix:

Within the _/tests/_ directory, create a _.pylintrc_ file and add this code to it:

```
[MASTER]
init-hook='import sys; sys.path.append("../src/")

```
When Pylint _sees_ a local .pylintrc it will execute it first and, in this case, add the source directory to the path. Pylint will otherwise ignore _sys.path.append()_ done from the file it is evaluation.

Regards,

Luis

FreemaGw:
Oh, thats interesting and good to know. Thanks a lot for fixing my problem (and future ones id suspect). Cheers.

PR #637, opened 2020-03-23
submission of L03

cjfortu:


ldconejo:
Hi Clemente,

Your implementation looks good. Unit testing however has multiple failures, while integration testing is passing. Could you take a look at that?

Thanks,

Luis

PR #652, opened 2020-03-31
Adjustment of L05

cjfortu:
Changed **test_clear_data()** to account for the fact that each user will have their own mongo databases loaded on their system.

The new test checks to make sure the **products_database** is no longer there, rather than checking to see if a specific list of databases is now there.


ldconejo:
Looks good now, thanks for taking the time to fix the findings from the first review.

-Luis

PR #653, opened 2020-04-02
Assignment 5

amax1213:


ldconejo:
Hi Allen,

The way _test_database.py_ is currently coded, you tests will fail.

The first issue is that you are combining Unix-style  and Windows directory routes. For example, on line 11:

`with open('./tests/test_products.csv', 'w') as file:`

While on line 33 you have this:

`counts, errors = db.import_data('.\\tests', 'test_products', 'test_customers',`

So, depending on whether you're running on a Windows system or a Unix-like system (Mac OS or Linux), your CSV files could not be created or your data could not be imported.

The second issue is that all your collections in _test_database.py_ use the _test__ suffix, while _database.py_ doesn't, so whenever you call _show_available_products_ or _show_rentals_ your tests will either fail if the collections used by _database.py_ are empty or have other data or falsely pass if the data matches what you're expecting.

One fix to this is to have _test_database.py_ use the same collection names (i.e., without the _test__ suffix) or make the collection names configurable via input parameter in _database.py_.

Once you fix those issues, your coverage will be at 97% and your assignment will be done.

-Luis

PR #655, opened 2020-04-02
first draft

FreemaGw:
Couldn't find any other instances where I would be acessing a whole list instead of just picking out a single entry, so I didn't know where else putting iterators into this code would be useful.


ldconejo:
Hi Ben,

Looks good. The idea is to make you aware of ways to use these constructs, the volume of data in this example is not huge, but it's good to know that the option of using an iterator is there.

I'll close this pull request for now, please reopen it once you have submitted your assignment on the course website.

-Luis

PR #661, opened 2020-04-05
Assignment 5 update

amax1213:


students/allen_maxwell/Lesson_05/database.py
```diff
@@ -0,0 +1,104 @@
...
+    return count, errors
+
+def show_available_products():
+    '''Returns a Python dictionary of products listed as available'''
+    results = {}
+    mongo = MongoDBConnection()
+    with mongo:
+        database = mongo.connection.hp_norton
+        LOGGER.info('Searching for available products')
+        for product in database.products.find({'quantity_available': {'$gt': '0'}}):
```
ldconejo, line 80: This line still uses "products" as the name of the collection, which will cause _test_show_available_products()_ to fail, because it's populating "test_products". Note that this line could also be:

`for product in database['products'].find({'quantity_available': {'$gt': '0'}}):`

Which would allow you to pass the name of the collection as a parameter (replacing 'products'). This is probably passing on your side because your _products_ collection already has the same entries that your tests are looking for.

students/allen_maxwell/Lesson_05/database.py
```diff
@@ -0,0 +1,104 @@
...
+    rented products matching the product id
+    '''
+    results = {}
+    mongo = MongoDBConnection()
+    with mongo:
+        database = mongo.connection.hp_norton
+        LOGGER.info('Searching for product_id: %s', product_id)
+        for rental in database.rentals.find({'product_id': product_id}):
+            for customer in database.customers.find({'user_id': rental['user_id']}):
+                results[customer['user_id']] = {
```
ldconejo, line 99: Similar issue here, you could replace _database.rentals_ with _database['rentals']_ and _database.customers_ with _database['customers']_ then make it configurable so that _show_rentals_ instead uses _database['test_rentals']_ and _database['test_customers']_ for your tests.

ldconejo:
Hi Allen,

See my comments on your code for some changes that are required. To avoid a false passing result from your tests, I'd recommend you drop (delete) the _customers_, _rentals_ and _products_ collections from _hp_norton_. 

You can do this from the Mongo CLI directly. 

- From a Linux/Mac terminal, type _mongo_ and hit enter.
- Type _use hp_norton_
- Type db.dropDatabase()

That way, your tests will run with a clean database. You could add code to the same from Python, just be careful (in real life) if actual collections and your test collections live within the same database.

Regards,

Luis

PR #664, opened 2020-04-07
submission of L07

cjfortu:


students/cjfortu/L07/assignment/database_multithread_join.py
```diff
@@ -0,0 +1,310 @@
...
+        else:
+            collection.insert_many(extracted_data)
+            LOGGER.info(f'{collection} successfully added.')
+            document_counts[i] = collection.count_documents({})
+
+    bad_file_path = False
+    for i in range(len(file_paths)):
+        thread = threading.Thread(target=write_data, args=(file_paths[i], collections[i], i))
+        thread.start()
+        thread.join()
```
ldconejo, line 189: If you run your _join()_ immediately after launching a thread or multiprocess, you essentially force it to be sequential. This is because _join()_ will sit there and wait for the thread to finish. Instead, do something like this:
   
```
    thread_list = []
    for i in range(len(file_paths)):
        thread = threading.Thread(target=write_data, args=(file_paths[i], collections[i], i))
        thread.start()
        thread_list.append(thread)

    for thread in thread_list():
        thread.join()
```

This will launch all threads, let them run and then wait for each one to finish. You will need to use queues for each thread to report back and consider that fact that threads might put the results in the queue in a different order from the order in which they were launched.

This will also work for multiprocessing, consider that you need to use _multiprocessing.queue_, other than that, the implementation should be identical to threading.

I'm wondering why your multiprocessing was so much faster, even with you joining each process right away. My guess is that a dedicated core still executes faster than one that is already dealing with other operations.

ldconejo:
Hi Clemente,

See my comments on your code. Could you collect the data again for threading and multiprocessing once you've separated the joining part into a different for loop? Let me know if you have any questions.

Thanks,

Luis

cjfortu:
Thanks for letting me know about that.  I also ended up finding out more about what causes issues using queues as well.

https://github.com/UWPCE-PythonCert-ClassRepos/SP_Python220B_2019/pull/666 <https://github.com/UWPCE-PythonCert-ClassRepos/SP_Python220B_2019/pull/666>

Clemente

PR #668, opened 2020-04-12
Lesson 9 submission

nullconfig:


ldconejo:
Hi Billy,

Your submission doesn't include _mongo_connect_ which is included in the imports and which I suppose includes the implementation of the database as a context manager. Also, I don't see any files for the recursion part of the assignment. Could you take a look and resubmit?

Thanks,

Luis

PR #672, opened 2020-04-13
lesson 09 submission

cjfortu:


ldconejo:
Hi Clemente,

Your implementation for the png finder clearly works, although it's a bit difficult to follow at times. The idea of recursion is a bit simpler.

For example:

```
def recursive_function(directory, images_found = []):
    list_of_elements = list_elements(directory)
    for element in list_of_elements:
        if is_image(element):
            images_found.append(element)
        if is_directory(element):
            recursive_function(element, images_found)
```

This is pseudo-code, meaning that this still would need additional coding (for example, creating a _is_image()_ function), but the core idea is there, in that it allows _recursive_function_ to call itself, passing a reference to the _images_found_ list, so that every image that is found by any of the calls to _recursive_function_ ends up there.

Good luck with assignment 10!

Luis

PR #673, opened 2020-04-13
lesson5

FreemaGw:
My log files are not working anymore, They are being created but not populated with any log messages, how do i fix this? Thanks!

### DIFF
--- a/students/Ben_freeman/lesson05/test.py
+++ b/students/Ben_freeman/lesson05/test.py
@@ -1,0 +1,30 @@
+# pylint: disable=unused-import, unused-wildcard-import,
+# pylint: disable=too-few-public-methods, too-many-arguments, wildcard-import
+# pylint: disable=logging-format-interpolation, pointless-string-statement
+import sys
+import unittest
+import database as db
+from unittest.mock import patch
+
+
+class TestFunctions(unittest.TestCase):
+
+    def setUP(self):
+        with patch("builtins.input", side_effect="yes"):
+            db.import_data("data_files", "products", "customers", "rentals")
+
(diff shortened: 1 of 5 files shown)


students/Ben_freeman/lesson05/test.py
```diff
@@ -0,0 +1,30 @@
...
+import sys
+import unittest
+import database as db
+from unittest.mock import patch
+
+
+class TestFunctions(unittest.TestCase):
+
+    def setUP(self):
+        with patch("builtins.input", side_effect="yes"):
```
ldconejo, line 13: Check this code and make sure that's really doing what you think. _side_effect_ is treated like an iterable, which means that in this case, since you're passing it a string, the first call to _input()_ will return "y", the next one "e" and then "s". So you're not really clearing the database. This line should look like this:

`with patch("builtins.input", side_effect=("yes",)):`

students/Ben_freeman/lesson05/test.py
```diff
@@ -0,0 +1,30 @@
...
+            db.import_data("data_files", "products", "customers", "rentals")
+
+    def test_import_data(self):
+        with patch("builtins.input", side_effect="yes"):
```
ldconejo, line 17: Same as above, there will be two calls to input, so in this case, this line should be:

`with patch("builtins.input", side_effect=("yes", "yes")):`

Or you could use _Mock()_ and set _input_ to always return "yes", no matter the number of calls. Either way works.

ldconejo:
Hi Ben,

Look at my comments on your test code, make sure all tests are passing after fixing that.

Thanks,

Luis

PR #684, opened 2020-04-20
assignment for lesson02

zyang2020:
file charges_calc.py is the updated version of old charges_calc.py
file debugger_log.txt is my Python debugger logfile
file charges_calc.log is the logfile created by running my updated script on the source data
file output.json is the output JSON file created by the updated script


ldconejo:
Hi Zhen,

Implementation looks good, could you run Pylint on your code and address any warnings? It's ok if you want to disable C0103 and W1202.

Please submit a new pull request with the updates.

Thanks,

Luis

PR #687, opened 2020-04-21
run pylint and fix the errors for lesson02 charges_calc.py

zyang2020:


zyang2020:
For lesson03 assignment, 
1. test_basic_operations.py is the unit test file for basic_operations.py
2. basic_operations.py contains five functions required by the assignment.
3. customer_model.py contains the Customer class (module) required by the assignement.
4. customer.db is the SQLite database.

PR #688, opened 2020-04-23
Attempt 1

FreemaGw:
One question that would be cool to know, is why my logging seems a bit screwy.

When I run poor_perf.py first, the log file is created along with all of the contents. If I then run good_perf.py, a new (and correct) log file is created, but not populated with log messages. Instead the poor_perf log file will receive the messages. If I run good perf first the reverse happens. I can avoid this by quiting python and rerunning, but this is annoying so I would like to avoid it. 

Thanks


ldconejo:
Hi Ben,

For the question on the log files, I assume you're running both scripts from the Python console. If that's the case, close the file handler at the end of both _poor_perf.py_ and _good_perf.py_ by calling:

`FILE_HANDLER.close()`

Also, the code could still be improved for performance. Consider the fact that the _analyze()_ function is opening the same file twice. You could instead examine the same line for the year and the presence of "ao" opening the file just one. Give that a try.

Regards,

Luis

PR #704, opened 2020-05-07
x

scotchwsplenda:


students/ScotchWSplenda/lesson01/assignment/inventory_management/main.py
```diff
@@ -0,0 +1,89 @@
...
+        print("item not found in inventory")
+
+
+def exitprogram():
+    '''some stuff16'''
+    sys.exit()
+
+
+if __name__ == '__main__':
+    '''some stuff17'''
```
ldconejo, line 84: Change this one to a regular comment:

`# some stuff17`

The _main_ block is just an _if_, instead of a fully fledged function, so this line is not considered a docstring but a regular string (as Pylint would put it, a _pointless_ string).

-Luis

PR #706, opened 2020-05-07
Adding files for lesson 3 "HP Norton customer database"

nskvarch:
pylint does not like the asterisk imports but trying to resolve from peewee is a challenge.


ldconejo:
Hi Niels,

Code looks good. For testing, coverage is at 81%, versus the 90% target. Try to add exception coverage, which should be an interesting exercise in how to trigger exceptions in unittest.

Regard,s

Luis

PR #709, opened 2020-05-11
I have a (dumb) question

FreemaGw:
I'm getting a pickle error when i Run my import function from "parallel.py". As far as I understand this means I have a lock object, and I'm pretty sure that it has to do with how im handling Mongodb, could you confirm if im on the right track? 

Cheers, 
thanks


ldconejo:
It's not dumb. It's actually a really interesting question, I ran into this one recently while designing new assignments for this course. 

If you're doing multiprocessing, each process will needs it's own connection to MongoDB (never mind the fact that you're probably running on localhost :-) ). If you want to use multiprocessing, which is not a bad strategy, you will need to modify _import_data_pre_function()_ to create its own instance of _MongoDBConnection()_.

Pretty cool stuff.

-Luis

PR #712, opened 2020-05-15
New question

FreemaGw:
First off, thanks for the tip on the db code, I got it working now! 

Second my code fails when I make the number of records too high (~400,000) and I don't know why.

Third the assignment page says "To show you have thought through your design, create and provide an example of where the program fails due to contention and explain why in code comments, and how that will be avoided when the system is running."
I don't know what contention means in this context, can you explain please?

Thanks,

Ben


ldconejo:
Hi Ben,

How do I trigger the error you see with n=400000? Right now, I get a similar error with emails not matching the key requirements, regardless of the number of records I use:

`bson.errors.InvalidDocument: key ' elisa.miles@yahoo.com' must not contain '.'`

Your is pretty cool and easy to run repeatedly, in that it allows you to clean up the database before every run.

Don't worry about the contention requirements. You're using multiprocessing, each process is running in a fully independent thread.

Luis

FreemaGw:
I must have goofed something in my last round of fiddling with the excel files, since the error you mentioned was the error I was getting, and it only happens on the ones that i modified. That said I am getting a weird error when I try to run my file more than once per Ipython instance. 

> `AttributeError: module '__main__' has no attribute '__spec__'

I can solve it by just quiting in between runs, but its obviously not ideal. I can google a fix for it, but I am curious why its happening.

PR #714, opened 2020-05-15
x

scotchwsplenda:


students/ScotchWSplenda/lesson01/assignment/inventory_management/main.py
```diff
@@ -0,0 +1,90 @@
...
+        print("1. add a new item to the inventory")
+        print("2. get item information")
+        print("q. quit")
+        user_prompt = input(">")
+    return valid_prompts.get(user_prompt)
+
+
+def get_price(item_code):
+    '''some stuff13'''
+    return(f"get price for {item_code}")
```
ldconejo, line 31: This works, but you don't need parenthesis in a return statement, so it could just be:

`return f"get price for {item_code}"`

PR #732, opened 2020-05-26
having trouble with append on the pngdiscover file

FreemaGw:
So I'm currently messing around with the last part of the assignment using ".py" in place of ".png". I think I'm like 90% of the way there, however I don't know how to get my dive function to properly append my big_list variable. If I put the print statement inside dive it works properly, but also prints each time it runs which is obviously a problem for a recursive function. I figured this would be easy to fix by just making sure that dive knows that big_list is a global variable (or non local when I was trying with closures) but this doesnt seem to be the case either. Any insight you can share would be appreciated.

Cheers,
Ben


students/Ben_freeman/lesson09/pngdiscover.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+
+big_list = []
+
+
+def dive(directory):
```
ldconejo, line 7: Pass _big_list_ as a parameter. Because of how lists work in Python, you will not be passing a copy of big_list, but a pointer to it (i.e., your function will be appending file names to the original list):

`def dive(directory, big_list)`

students/Ben_freeman/lesson09/pngdiscover.py
```diff
@@ -0,0 +1,22 @@
...
+    global big_list
```
ldconejo, line 8: Making _big_list_ a global is not necessary, as I said, it's the same list everywhere.

students/Ben_freeman/lesson09/pngdiscover.py
```diff
@@ -0,0 +1,22 @@
...
+    contents = os.listdir(directory)
+    picture_holder = []
```
ldconejo, line 10: It's ok to use _picture_holder_ if for some reason you want to know how many files you have found in this particular run, but you could just append files straight into _big_list_

students/Ben_freeman/lesson09/pngdiscover.py
```diff
@@ -0,0 +1,22 @@
...
+    for items in contents:
+        #print(items)
+        try:
+            dive(items)
+        except:
+            if ".py" in items:
```
ldconejo, line 16: Keep in mind that you're looking for _.png_ files.

students/Ben_freeman/lesson09/pngdiscover.py
```diff
@@ -0,0 +1,22 @@
...
+                picture_holder.append(items)
+    print(picture_holder)
+    big_list.append(picture_holder)
```
ldconejo, line 19: If you use _append()_ the entire _picture_holder_ list will be added as a **single element** into big list. You probably want to use _extend()_ that will add each element in _picture_holder_ as a separate element into _big_list_

ldconejo:
Hi Ben,

Looks like you're on the right track. See my comments on your code regarding your specific question and some other potential changes.

Regards,

Luis

PR #737, opened 2020-05-27
Attempt#1

FreemaGw:
Hi, I think I have everything working. However I have one question and one clarification.

Question: How do you use the "@decorator" style of using decorators in a case when you have multiple input variables like in the case of my lesson02 update?

Clarification: In my previous push, you suggested: "If you use append() the entire picture_holder list will be added as a single element into big list. You probably want to use extend() that will add each element in picture_holder as a separate element into big_list"

however on the assignment page the format shown is: [“full/path/to/files”, [“file1.png”, “file2.png”,…], “another/path”,[], etc]

I can do it either way, just wanna know if I should group each png found together with the other ones in the same directory or if I should make each one stand alone.

Thanks! Also have a nice vacation if I don't submit my lesson 10 before the 1st.


ldconejo:
> Hi, I think I have everything working. However I have one question and one clarification.
> 
> Question: How do you use the "@decorator" style of using decorators in a case when you have multiple input variables like in the case of my lesson02 update?
> 
> Clarification: In my previous push, you suggested: "If you use append() the entire picture_holder list will be added as a single element into big list. You probably want to use extend() that will add each element in picture_holder as a separate element into big_list"
> 
> however on the assignment page the format shown is: [“full/path/to/files”, [“file1.png”, “file2.png”,…], “another/path”,[], etc]
> 
> I can do it either way, just wanna know if I should group each png found together with the other ones in the same directory or if I should make each one stand alone.
> 
> Thanks! Also have a nice vacation if I don't submit my lesson 10 before the 1st.

Hi Ben,

About decorators in functions with multiple input parameters, I take it you mean cases in which the function being decorated takes multiple parameters. Take a look at this code and let me know if it answers your question:

```
def printing_decorator(func):
    def wrapper(*args, **kwargs):
        list_output = func(*args, **kwargs)
        return f"My favorite dessert is {list_output[0]} {list_output[1]} {list_output[2]}"
    return wrapper

@printing_decorator
def turn_into_list(color, fruit, dish):
    return [color, fruit, dish]

if __name__ == '__main__':
    print(turn_into_list("blue", "berry", "muffin"))
```

On the list of files for the recursive function, I forgot that the assignment requires it like that, for some reason I was expecting it to simply require returning a list of files found.

I'm going to close this pull request for now. Please open a new one once the decorator part is ready.

-Luis

PR #747, opened 2020-05-31
Finished Lesson03

mgummel25:
Added Lesson03 files.


ldconejo:
Good job! Your code also forced me to update peewee, I was running an older version :-)

-Luis

PR #748, opened 2020-06-01
Submiting Lession9

Reem0330:
Thank you for reviewing my assignment


ldconejo:
Your implementation of the decorator actually solved a question from another student on how to pass the debug level as a parameter. It was really nice coding.

-Luis

PR #749, opened 2020-06-01
Please review

tianxie812:
Hi Luis,

Per our conversation, I have created a pull request for your review.

Thanks,

Tian

### DIFF
--- a/students/Tianx/Lesson1/Assignment/inventory_management/main.py
+++ b/students/Tianx/Lesson1/Assignment/inventory_management/main.py
@@ -7,6 +7,7 @@ import electric_appliances_class
 
 FULL_INVENTORY = {}
 
+
 def main_menu(user_prompt=None):
     """
     :param user_prompt:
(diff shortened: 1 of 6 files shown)


students/Tianx/Lesson1/Assignment/inventory_management/main.py
```diff
@@ -0,0 +1,106 @@
...
+    Function to get item info
+    """
+    item_code = input("Enter item code: ")
+    if item_code in FULL_INVENTORY:
+        print_dict = FULL_INVENTORY[item_code]
+        for key, value in print_dict.items():
+            print("{}:{}".format(key, value))
+    else:
+        print("Item not found in inventory")
+
```
ldconejo, line 93: Hi Tian,

Good catch. This function should either return _True_ if the item code is found and _False_ if it's not. Please go ahead and do the change here and on the test (you should check both item found and not found).

Regards,

Luis

PR #754, opened 2020-06-03
l3

scotchwsplenda:
test_sqlite: 
-doesnt work when i put it in the subfolder 'tests''
-if i indent the 'customers' dicts under def setup they dont get recognized
in the lower down methods, something to do with 'self'
-the db gets created without being explicitly told to do so

customer_model:
-Peewee ORM, sqlite and Python
-what is this 'peewee.' preamble? -> only if you do a * import
-class = table
-model = file where you are saving the classes/tables -> like a db essentially
-why don't i have to 'connect' or 'close'?


ldconejo:
> test_sqlite:
> -doesnt work when i put it in the subfolder 'tests''
> -if i indent the 'customers' dicts under def setup they dont get recognized
> in the lower down methods, something to do with 'self'
> -the db gets created without being explicitly told to do so
> 
> customer_model:
> -Peewee ORM, sqlite and Python
> -what is this 'peewee.' preamble? -> only if you do a * import
> -class = table
> -model = file where you are saving the classes/tables -> like a db essentially
> -why don't i have to 'connect' or 'close'?

Hi Oliver,

To your question, your test code would work with _test_sqlite.py_ in the _tests_ folder if you call it from the root folder:

`$ python -m coverage run --include=basic_opps_dict.py -m unittest tests/test_sqlite.py`

The issue is paths. Your code will try to import from the same folder in which it resides and then in the Python path, since _basic_opps_dict.py_ and _customer_model.py_ are not there, your testing will fail if you attempt to run it from the _tests_ directory. Another choice would be to have your test code use _sys.path.append()_ to add the path where your target code is located.

_> -if i indent the 'customers' dicts under def setup they dont get recognized
> in the lower down methods, something to do with 'self'_

Could you provide more details on this? _self_ is simply how an instance of a class refers to itself within the class.

`> -the db gets created without being explicitly told to do so`

Not really, when you import _customer_model.py_ the following line runs:

_db = SqliteDatabase('norton.db')_

_basic_opps_dict.py_ also runs these:

_Customer.drop_table()
Customer.create_table()_

Those events effectively create the database.

_> -what is this 'peewee.' preamble? -> only if you do a * import_

If you only do _import peewee_ all functions within the peewee modules have to be prefixed with _peewee_. Same goes for any other Python module.

_> -why don't i have to 'connect' or 'close'?_

This one baffled me a bit. It turns out Peewee doesn't really require you to connect:

http://docs.peewee-orm.com/en/3.6.0/peewee/database.html#connection-management

_it is not necessary to explicitly connect to the database before using it_

I look forward to your next assignment,

Luis

PR #762, opened 2020-06-08
lesson 2

ikempcode33:
Finished lesson 2. I think I understand when to add in logging to track the program but I am still confused on the best practices with logging. Do you have any tips to best use logging in general or any other readings or example code you can point me to for best logging practices? Did I use it correctly in this instance? Thank you!


students/ikempcode33/Lesson 2/charges_calc.py
```diff
@@ -0,0 +1,115 @@
...
+    logging.error("Calculating additional fields.")
+    for value in data.values():
+        try:
+            rental_start = datetime.datetime.strptime(
+                value['rental_start'], '%m/%d/%y')
+            rental_end = datetime.datetime.strptime(
+                value['rental_end'], '%m/%d/%y')
+        except KeyError:
+            value['total_days'] = (rental_end - rental_start).days
+            continue
```
ldconejo, line 81: This _continue_ statement is causing code in lines 82-92 to be unreachable. Take a look and try to fix it, I would think the indentation is wrong.

students/ikempcode33/Lesson 2/charges_calc.py
```diff
@@ -0,0 +1,115 @@
...
+    file_handler.setLevel(debug_level)
+    console_handler = logging.StreamHandler()
+    console_handler.setLevel(logging.DEBUG)
+    console_handler.setFormatter(formatter)
+    logger = logging.getLogger()
+    logger.setLevel(debug_level)
+    logger.addHandler(file_handler)
+    logger.addHandler(console_handler)
+
+    loggingSetup = "Logging set up complete."
```
ldconejo, line 41: Just to respect PEP8 convention, this one should be called _logging_setup_ (snake case versus camel case).

ldconejo:
Hi Isabella,

Take a look at a couple of comments for things that you need to correct, code wise. Your implementation of logging is good; you should consider adding warnings, for example, for issues in the source data that your code can overcome but indicate that something might be wrong in that source data. 

Errors should be used for things that would make your code crash if you didn't have exception handling for it (for example, a division by zero). Right now, you have "Calculating additional fields" as en error, which should not be the case.

Your logging should also provide useful details on why something is being raised as a warning or an error (what line in the input data, what was the issue, how's your code handling it?).

Note that in this course you can always resubmit assignments to fix any issues detected and still get the full grade, even if you have to resubmit multiple times.

Regards,

Luis

PR #763, opened 2020-06-09
ganyoa: commit lesson02 files

ganyoa:
Hi, please see attached lesson two assignment. I struggled finding a good way to close out the calculate_additional_fields function. Is there a good catch-all 'except'...without a blanket Error?  Thank you


ldconejo:
Excellent, it's been quite a while since I've seen a flawless assignment for lesson 2.

-Luis

ldconejo:
> Hi, please see attached lesson two assignment. I struggled finding a good way to close out the calculate_additional_fields function. Is there a good catch-all 'except'...without a blanket Error? Thank you

I forgot to reply to your question: In my _personal_ opinion as a developer, there's no good catch-all exception. You can group multiple exception types in a single except, assuming they can all be handled in the same way. 

On the other hand, using a bare except is never worth the risk. The exception you didn't see during development, the one that will only occur on December 31st, when you want to be with family and friends and not working, and that should have a different handling, will get caught in the catch-all and you'll be debugging for hours trying to figure out why your code is not working but not crashing.

Just my two cents (I really don't like bare exceptions :-) ).

-Luis

PR #765, opened 2020-06-09
charges_calc.py

ikempcode33:
I fixed some of the comments you made. However, I tried moving the except Key error and the continue statement and I am getting what seems to be the right output but I think it is still not working properly. If I remove the continue statement it seems incorrect because the code underneath would only be executed if a key error was present.  Anyways, can you take a look and tell me where my try excepts may be going wrong? I also put the except key error below the if statement on line 88 but then was returning another value error, but with it where it is not it runs without any errors on my side.


ikempcode33:
Actually look at my latest one, ignore the previous one with the previous comment. This one seems to work better with how the try excepts are set up

students/ikempcode33/Lesson 2/charges_calc.py
```diff
@@ -0,0 +1,129 @@
...
+def calculate_additional_fields(data):
+    """calculate fields for output data"""
+    logger.info("Calculating additional fields.")
+    for value in data.values():
+        try:
+            rental_start = datetime.datetime.strptime(
+                value['rental_start'], '%m/%d/%y')
+            rental_end = datetime.datetime.strptime(
+                value['rental_end'], '%m/%d/%y')
+        except:
```
ldconejo, line 83: Never use bare exceptions (also known as _catch-all_ exceptions), after you release your code, some new exception type will hit your code, get caught in the catch all and have you debugging late on a Sunday night, while you're on vacation, trying to figure out why your code is neither working nor crashing. Guaranteed :-)

In this case, the exception type should be _ValueError_

students/ikempcode33/Lesson 2/charges_calc.py
```diff
@@ -0,0 +1,129 @@
...
+            logger.error("Bad date, rental_start or rental_end: %s", value)
+            continue
+        try:
+            value['total_days'] = (rental_end - rental_start).days
+            if value['total days'] < 1:
+                logger.warning("start date after end date: %s", value)
+        except KeyError:
+            logger.debug("KeyError always generated on Value[] inserts?")
```
ldconejo, line 91: On line 88, you have this:

`if value['total days'] < 1:`

Which is not the same as this:

`if value['total_days'] < 1:`

It took me a while to notice it :-)

ldconejo:
Almost there, take a look at my comments, couple of easy fixes, open a new pull request and you're good to go with this one.

-Luis

PR #768, opened 2020-06-10
fixing debug level setup issue with int

tianxie812:


ldconejo:
Looks perfect now!

-Luis

PR #769, opened 2020-06-12
sp_py220 lesson3 assignment relational db

mmanc125uw:


ldconejo:
Coverage is below 90% target, at 89%. It's not a big deal, but I think it's a good opportunity for practicing exception coverage.

Currently, the _IntegrityError_ trap on line 57 is not covered, same as the _DoesNotExist_ trap on line 110. Either of those (or both) would get your coverage higher than 90% and provide the change of triggering an exception with your tests (let me know if you need help on that one, note that testing for an exception should not make your code crash).

-Luis

mmanc125uw:
Hi Luis
I made the changes per your suggestion and coverage up to 97%.   I included a readme, and coverage_results.txt for your reference.  Below is the pull request I submmitted.

https://github.com/UWPCE-PythonCert-ClassRepos/SP_Python220B_2019/pull/775


Thanks
Mike

PR #771, opened 2020-06-12
x

scotchwsplenda:
I need help figuring out how to make " sys.path.append() " work to save my test in a separate folder.


ldconejo:
Hi Oliver,

Code looks good, coverage is lower than target at 73% versus 90%. Also, make sure you run Pylint on your code and address any warnings there to get to 10/10.

For coverage, adding test coverage to your exceptions should do the trick.

What issues are you having with _sys.path.append()_?

Regards,

Luis

PR #772, opened 2020-06-13
Adding activity and assignment 1, 10/10 pylint, 92percent cov on main

mokiko81:
Hello,

  Submitting assignment 1 in lesson 01 for review:

- I have 10/10 pylint coverage when disabling the following on the pylintrc file:
     1) **unused argument** - for the argument in market_prices.py as it said not to change the file and only to mock its output (one thing I was unsure of is if I should modify it to produce an error when called, it is mocked in all tests as 'FakeOutput')
    2) **too-few-public-methods**
    3) **too many arguments** - I could clean this up with some sort of *args solution but didn't think it necessary

- My test_unit.py has 100% coverage on all files, 92% on main.py
- My test_integration.py file runs through each of the 3 types of inventory items, asserts after each function call to look at the MAIN_INVENTORY dictionary and then looks one of them up in item_info. Is this enough for the integration test or should it have more?

Thank you!

### DIFF
--- a/students/adam_strong/lesson01/activity/README.md
+++ b/students/adam_strong/lesson01/activity/README.md
@@ -1,0 +1,32 @@
+# Calculator
+
+## Instructions
+
+Your assignment is to complete testing and linting on the calculator from the lesson videos.
+
+There's one new addition since the videos: I've separated the unit tests and the integration tests into two separate test files.
+
+## Your Goals
+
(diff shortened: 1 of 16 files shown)


ldconejo:
Perfect first submission and good attention to detail. Keep up the good work!

-Luis

ldconejo:
Hi Adam,

Could you submit this assignment on open edX, so that I can grade it? I'll close this pull request for now, just reopen it once you have submitted the assignment on the course website.

Thanks,

Luis

PR #773, opened 2020-06-14
attempt1

FreemaGw:
Decorators count as meta programming right? It seemed like the most straight forward way to go about this. If not I figured out how to do it with classes but its kind of clunky and not aesthetically pleasing in my opinion. Cheers!


ldconejo:
Using decorators is fine. Data looks good, congratulations on completing the last assignment!

-Luis

FreemaGw:
Thank you! You were a very helpful instructor!

PR #774, opened 2020-06-14
lesson 3

ikempcode33:
Still having trouble getting add_customer test to pass due to an assertion error of the decimal place... was unable to fix it properly by altering the number, etc.


students/ikempcode33/Lesson 3/test_basic_operations.py
```diff
@@ -0,0 +1,85 @@
...
+        logger.info('creating tables')
+        database.drop_tables([Customer])
+        database.create_tables([Customer])
+        logger.info('tables are created.')
+
+        logger.info('Adding data to database')
+        add_customer(1773, 'Karen', 'Smith',
+                     '142 42 st, Bellevue, WA 98105',
+                     '123-456-7890', 'ksmith99@gmail.com',
+                     40000.00, True)
```
ldconejo: These are the last two parameters _add_customer()_ expects:

`status, credit_limit`

There's a problem with the order of the parameters in your test.

ikempcode33:
Fixed that, not sure why i didn't notice it. Thank you!

ldconejo:
> Fixed that, not sure why i didn't notice it. Thank you!

Sometimes you just need another pair of eyes to look at your code. Happens to me all the time :-)

PR #776, opened 2020-06-16
Help, pylint disable doesn't work

tianxie812:
Hi Luis,

For some reason I was not able to disable pylint errors. Most of them are import errors which is very strange. Would you please take a look at it?

Thanks,
Tian

### DIFF
--- a/students/Tianx/Lesson3/src/basic_operations.py
+++ b/students/Tianx/Lesson3/src/basic_operations.py
@@ -1,27 +1,112 @@
+import logging
 from peewee import *
+from customer_model import Customer
 
-# add_customer(customer_id, name, lastname, home_address, phone_number, email_address, status, credit_limit): This function will add a new customer to the sqlite3 database.
-#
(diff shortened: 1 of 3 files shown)


students/Tianx/Lesson3/src/basic_operations.py
```diff
@@ -1,27 +1,112 @@
...
-# update_customer_credit(customer_id, credit_limit): This function will search an existing customer by customer_id and update their credit limit or raise a ValueError exception if the customer does not exist.
-#
-# list_active_customers(): This function will return an integer with the number of customers whose status is currently active.
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
 
+db = SqliteDatabase('customers.db')
+db.create_tables([Customer])
+
+# pylint: disable = W0614, W0401, C0301, C0114, W1203
```
ldconejo, line 11: The disable only applies to code lines after the location of the disable, try moving it to line 1 and run Pylint again.

ldconejo:
Hi Tian,

See my comment on where to place the pylint:disable comment.

-Luis

PR #780, opened 2020-06-18
Submitting assignment 3

tianxie812:


ldconejo:
Hi Tian,

Your implementation looks good, test coverage is lower than target at 82% versus 90% target. Could you add test coverage to your exceptions? That should get you all the way to 100%.

-Luis

PR #781, opened 2020-06-18
completed lesson06

mgummel25:


ldconejo:
Excellent analysis and performance improvement. Keep it up, you're very close now!

-Luis

PR #782, opened 2020-06-18
Submitting Lesson02 for review

mokiko81:
Hello,

  This is the debugging/logging lesson 02 assignment for review. Not sure why it also added my lesson01 materials which I didn't change at all.

For the lesson02 assignment:
- I have 10/10 pylint coverage
- I added the final log file from running the finished script
- I added in a separate folder the log files that were generated before the final script
- I added the work I did with the debugger with a brief description of what I was doing/looking for during the debugging


students/adam_strong/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,146 @@
...
+                        3=Error messages, warnings and debug messages')
+    try:
+        return parser.parse_args()
+    except TypeError:
+        sys.exit()
+
+def set_logger(debug):
+    '''Sets the logging level input from command line'''
+    if debug > 0:
+        if debug == 1:
```
ldconejo, line 46: There is one thing missing here: As coded, this will also accept any invalid value (such as -d 10). Try to add a condition to error out if the user enters an invalid debug level.

ldconejo:
Implementation looks good, see my comments on your pull request one detail to be fixed. Submit a new pull request once that's fixed and I'll update the grade.

Regards,

Luis

PR #792, opened 2020-06-24
lesson 5

ikempcode33:
testing the available products is not working for some reason. The other two tests run. There must be some kind of issue with reading in that csv file but i cannot pinpoint it. CSV files are being taken in correctly.


students/ikempcode33/lesson 5/database.py
```diff
@@ -0,0 +1,99 @@
...
+def show_available_products():
+    """Returns a dictionary of available products"""
+    mongo = MongoDBConnection()
+    with mongo:
+        db = mongo.connection.media
+        products = db['products']
+        available_products = {}  #Dictionary output
+        # Iterate through products for dict
+        for item in products.find():
+            available_products[item['product_id']] = {'description': item['description'],
```
ldconejo, line 81: Note that in your database, you really don't have a _product_type_ or _quantity_ fields. Instead, you have this:

```
> db.products.find()
{ "_id" : ObjectId("5ec867e8ab332d42405e42b4"), "product_id" : "2001", "description" : "table_lamp", "type" : "electronics", "number_available" : "4" }
```
So, _product_type_ is just _type_ and _quantity_ is _number_available_. So, first step will be to correct that. At least on my side, I get other errors on the comparison, but let's tackle one thing at a time.

ldconejo:
Hi Isabella,

See me comments on your code for one fix to be made, let me know if you have any more questions!

-Luis

ikempcode33:
I tried changing it to just type and number_available but I still get the assertion error when testing it. Also it confuses me because in my CSV file I have them listed as product_type and quantity in the rows. 

 

 

 

----------------------------------------------------

Isabella Kemp 

Department of Bioengineering

Undergraduate Researcher

University of Washington Class of 2021

Ikemp33@uw.edu

PR #793, opened 2020-06-25
Adding lesson03 work for review - questions about pylint disables

mokiko81:
Adding both Lesson03 activity and Lesson03 assignment -

Feel the assignment is in a good place:
- Tests with full coverage of basic_operations.py (except for __name__ == '__main__' line)
- Pylint 10/10 with the following exceptions which I wanted to ask about:
> disabled (invalid-name) because it wants snake casing for my temporary variables of 'c' and 'e'
> disabled (wildcard-import and unused wildcard import) for customer_model.py
> disabled (not-an-iterable) for looping through Customers records in my list_all_customers() - This one I'm most unsure of, as I'm sure there is another way to do what I needed to do (format print out each record in Customers)
> disable (too-few-public-methods)

I used an AutoField for the PrimaryKey which is the customer_id (which I could change to something like email).
I made email unique to prevent duplicate records (therefore could be used as PrimaryKey)

One other question I had was on the assignment requirement to have some tests fail, I took this to mean to test some of the exceptions that could occur, I have these with Integrity Errors, peewee.DoesNotExist errors and ValueErrors that can be raised.

Thank you!
- Adam Strong

### DIFF
--- a/students/adam_strong/lesson03/assignment/basic_operations.py
+++ b/students/adam_strong/lesson03/assignment/basic_operations.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python
+"""
+Model to interact with customers.db
+for use by all departments at HP Norton
+pylintrc file: disabled - invalid-name(for snake casing), wildcard-import,
+unused-wildcard-import, too-few-public-methods, not-an-iterable
+"""
+
+import logging
+import pathlib
+import peewee
+from playhouse.shortcuts import model_to_dict
+from customer_model import *
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+logger.info('basic_operations.py is at work')
+
+FIRST_NAME = 0
+LAST_NAME = 1
+HOME_ADDRESS = 2
+PHONE_NUMBER = 3
+EMAIL_ADDRESS = 4
+IS_ACTIVE = 5
+CREDIT_LIMIT = 6
+
+REPORT_STRING = 'ID#{:5}:::{:20} P: {:9}, E: {:20}, Active:{:1}, Limit:{:8}'
+
+def add_customer(customer_tuple):
+    ''' This function will add a new customer to the sqlite3 database.'''
+    logger.info('In add customer function')
+
+    with database.transaction():
+        new_customer = Customers.create(first_name=customer_tuple[FIRST_NAME],
+                                        last_name=customer_tuple[LAST_NAME],
+                                        home_address=customer_tuple[HOME_ADDRESS],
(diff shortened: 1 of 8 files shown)


students/adam_strong/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,116 @@
...
+    query = Customers.select(fn.COUNT(Customers.customer_id)).where(Customers.is_active == 1)
+    count = query.scalar()
+    logger.info('There are a total of %s active customers', str(count))
+    return count
+
+
+def list_all_customers():
+    '''This function returns a list of lists of all the customers'''
+    cust_list = []
+    for c in Customers:
```
ldconejo, line 102: Remember _Customers_ is a SQL db table, so you could get a list of all customers by running a query::

`Customers.select()`

If you only wanted active customers:

`Customers.select().where(Customers.status == "active")`

ldconejo:
> Adding both Lesson03 activity and Lesson03 assignment -
> 
> Feel the assignment is in a good place:
> 
> * Tests with full coverage of basic_operations.py (except for **name** == '**main**' line)
> * Pylint 10/10 with the following exceptions which I wanted to ask about:
> 
> > disabled (invalid-name) because it wants snake casing for my temporary variables of 'c' and 'e'
> > disabled (wildcard-import and unused wildcard import) for customer_model.py
> > disabled (not-an-iterable) for looping through Customers records in my list_all_customers() - This one I'm most unsure of, as I'm sure there is another way to do what I needed to do (format print out each record in Customers)
> > disable (too-few-public-methods)
> 
> I used an AutoField for the PrimaryKey which is the customer_id (which I could change to something like email).
> I made email unique to prevent duplicate records (therefore could be used as PrimaryKey)
> 
> One other question I had was on the assignment requirement to have some tests fail, I took this to mean to test some of the exceptions that could occur, I have these with Integrity Errors, peewee.DoesNotExist errors and ValueErrors that can be raised.
> 
> Thank you!
> 
> * Adam Strong

Hi Adam,

You are on the right track. I added a comment on your function for listing all customers, I'd be better to use a query for that. I'm ok with your other rule disables, if possible, try to avoid single-letter variables, they're a pain if you have to return to your code later and figure out what it was doing. For example:

`for customer in Customers:`

Is more readable than:

`for c in Customers:`

Finally, wildcard imports are ok for assignments, they're not always a good idea for code you intend to release, as anything that gets imported takes up memory and makes your program unnecessary bloated (since you won't use most of the imports).

I look forward to your submission for this assignment.

-Luis

PR #796, opened 2020-06-27
Assignment 2: Logging & Debugging

stellie82:
- complete exercise for debugging
- update code to be able to input/output files
- update code to handle reversed start/end rental dates
- update code to pass on no units rented or no rental date
- add debug messages and logging to file


students/stellie/lesson02/assignment/charges_calc.py
```diff
@@ -0,0 +1,141 @@
...
+    console_handler.setFormatter(formatter)
+
+    logger = logging.getLogger()
+    log_levels = {0: logging.NOTSET, 1: logging.ERROR, 2: logging.WARNING,
+                  3: logging.DEBUG}
+    try:
+        logger.setLevel(log_levels[level])
+    except KeyError:
+        print('Debugging level is invalid.  Please enter a level between 0-3.')
+
```
ldconejo, line 51: Hi Stella,

Looks good. On this exception, force your code to exit, otherwise it will issue a warning on screen, which will not really be visible as it will be followed by log output. One option is to use _sys.exit()_ for it. You can submit an updated pull request with this change to get the full grade.

-Luis

PR #816, opened 2020-07-16
@contextmanager question

scotchwsplenda:
@contextmanager
def MongoDBConnection(host, port):
    '''can 'mongobaby' be replaced by 'connection' is 'connection'
    a keyword?'''
    mongobaby = MongoClient('127.0.0.1', 27017)
    return mongobaby
    yield
    mongobaby.close()


ldconejo:
> @contextmanager
> def MongoDBConnection(host, port):
> '''can 'mongobaby' be replaced by 'connection' is 'connection'
> a keyword?'''
> mongobaby = MongoClient('127.0.0.1', 27017)
> return mongobaby
> yield
> mongobaby.close()

Hi Oliver,

First of all, if pretty cool that you did not just keep the same implementation that you had for assignment 5, which was already implemented as a context manager:

```
class MongoDBConnection(): 

45    """MongoDB Connection""" 
46 
47    def __init__(self, host='127.0.0.1', port=27017): 
48        """ be sure to use the ip address not name for local windows""" 
49        self.host = host 
50        self.port = port 
51        self.connection = None 
52 
53    def __enter__(self): 
54        self.connection = MongoClient(self.host, self.port) 
55        return self 
56 
57    def __exit__(self, exc_type, exc_val, exc_tb): 
58        self.connection.close() 
```

That said, _connection_ is not a reserved word in Python, you can use it safely. It's still a good question to ask, as Python will actually let you redefine any reserved words without raising any errors. However, if you do:

```
user_selection = input("Select one option: ")
input = "This used to be a reserved word"
user_selection = input("Select another option: ")
```

Will crash on the third line, since _input()_ is not longer a function. 

Just in case, Pylint will warn you if you're redefining reserved words.

Regards,

Luis

PR #825, opened 2020-07-22
Lesson 07 submission for review

mokiko81:
I didn't submit this yet on the class site because I wasn't sure if the exercise was complete, but I will do that in the future when I believe the files are ready to submit (from your last email).

My big question about this exercise is the requirement to show where contention causes the program to fail, I had several versions of my parallel.py (which I kept in a /parallelization_attempts folder which I also have in this pull request). I am unsure if I answered that question sufficiently enough which is in my findings.txt document. Although, I did try different versions (some of which were worse than the linear.py - such as v1 using async/await and the motor api) I also tried different record amounts (~10, ~1000, ~10000) and saw a consistent 20% speed up for running all 3 imports in parallel (using threading).

Thank you!
  Adam Strong

### DIFF
--- a/students/adam_strong/lesson07/assignment/findings.txt
+++ b/students/adam_strong/lesson07/assignment/findings.txt
@@ -1,0 +1,150 @@
+
+### Just adding total import time to lesson 05 code ####
+
+In [1]: run linear.py                                                      
+2020-07-20 12:46:33,367 linear.py:66  INFO IMPORTING DATA
+2020-07-20 12:46:33,448 linear.py:97  INFO Added: [8, 10, 8]
+2020-07-20 12:46:33,449 linear.py:71  INFO Done with importing all 3 files.
+2020-07-20 12:46:33,449 linear.py:119 INFO Finished with the import - main loop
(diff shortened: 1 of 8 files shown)


ldconejo:
> I didn't submit this yet on the class site because I wasn't sure if the exercise was complete, but I will do that in the future when I believe the files are ready to submit (from your last email).
> 
> My big question about this exercise is the requirement to show where contention causes the program to fail, I had several versions of my parallel.py (which I kept in a /parallelization_attempts folder which I also have in this pull request). I am unsure if I answered that question sufficiently enough which is in my findings.txt document. Although, I did try different versions (some of which were worse than the linear.py - such as v1 using async/await and the motor api) I also tried different record amounts (~10, ~1000, ~10000) and saw a consistent 20% speed up for running all 3 imports in parallel (using threading).
> 
> Thank you!
> Adam Strong

Hi Adam,

This submission looks good. The requirement to make the program fail via contention is not particularly useful (at least not for the purposes of the assignment). Please go ahead and submit on the course website, so that I can grade it. Reopen the pull request once you do so.

Regards,

Luis

PR #833, opened 2020-07-29
Assignment 5: Consuming APIs with NoSQL

stellie82:
- setup connection to Mongo DB
- create function to create collection in database, import CSV file and upload file data to collection
- create functions to show all available products for rental (anything that does not have a zero value) and customer information for specific rentals (by product ID)
- add coverage testing for codebase


ldconejo:
Good job, coverage is at 98% and no issues reported by Pylint.

While testing on my system, your code crashed the first time because a previous review had left another "hp_norton" database that caused issues to yours. So one thing to consider is to also drop the entire database in your _clear_collections()_ function, which can be done with _database.dropDatabase()_. Nothing to fix in your, but something to consider if you run into a similar issue later on.

Keep up the good work!

Luis

stellie82:
@ldconejo that is good advice and I'll keep that in mind going forward, thanks!

PR #834, opened 2020-07-30
ganyoa: lesson04

ganyoa:
Hi, After I added the additional logging (console and file) I started seeing duplicate logs in the console. The db.log output file seems to be recording correctly, but I can't figure out why I see doubles on the console. I included a file (results_of_tests.txt) to show all my results including the dups in the unittest.  Thanks! Anthony Ganyo


ldconejo:
> Hi, After I added the additional logging (console and file) I started seeing duplicate logs in the console. The db.log output file seems to be recording correctly, but I can't figure out why I see doubles on the console. I included a file (results_of_tests.txt) to show all my results including the dups in the unittest. Thanks! Anthony Ganyo

Hi Anthony,

Looks good. On the duplicated console events, you ran into an interesting bug. Here's what happened:

**The short version:** You have log entries in _customer_model.py_ than run before you have setup your event handlers in _basic_operation.py_, triggering the creation of a duplicated handler (there's no limit to how many handlers of the same type you can have).

**The long version:**

- In _basic_operation.py_ you import a couple of elements from _customer_model.py_. The process of importing, even a partial import using _from_ like in this case, causes any code lines not inside of a function, class, etc., to run. This includes lines 4 and 5:

```
logging.basicConfig(level=logging.INFO, format="%(asctime)s %(filename)s:%(lineno)-3d %(levelname)s %(message)s")
logging.info('Define data (the schema) for Customer table in customers.db')
```

- Since the code above runs before you have setup your handlers, it creates its own, which goes to the console. When you then add the console handler, you end up with two different handlers outputting to the console.

Possible solutions: Just make sure that all log messages occur only after your handlers have been setup.

Hope this helps!

Luis

fd

PR #839, opened 2020-07-31
Lesson 09 assignments for review

mokiko81:
Hello,

 I did the 3 assignments for Lesson 09 and had a few questions (on parts 1 and 2) before I submit them through the website.

1) Adding decorators -
   I implemented a solution to this but I feel it is not very elegant and wanted to see if there is someway to improve it. I made a logging decorator called 'toggle_logging' as well as a new argparse argument called -toggle/-t with two valid options: ON/OFF (default is ON). The implementation works but I had difficulty having the decorator re-enable logging after the function has completed. Instead I have the re-enabling in the 'if __name__ == '__main__'' clause between the function calls. This seems very messy, I also used an implementation of the global keyword in my decorator which pylint doesn't like.

2) Context managers-
   Because my original Lesson 05 assignment already used a context manager for mongo I couldn't really figure out what other useful things to add. I decided to incorporate a print statement for the __init__, __enter__ and __exit__ functions within the MongoDBConnection class. The __init__ print statement seems the most useful as it prints out the port and host information when initialized.

3) Recursion-
  This one was fun, I don't really have any questions on it, it was a good assignment and I'm always curious of more real world implementations of the type of recursion that python can do.

Thank you!
  - Adam Strong


ldconejo:
> Hello,
> 
> I did the 3 assignments for Lesson 09 and had a few questions (on parts 1 and 2) before I submit them through the website.
> 
> 1. Adding decorators -
>    I implemented a solution to this but I feel it is not very elegant and wanted to see if there is someway to improve it. I made a logging decorator called 'toggle_logging' as well as a new argparse argument called -toggle/-t with two valid options: ON/OFF (default is ON). The implementation works but I had difficulty having the decorator re-enable logging after the function has completed. Instead I have the re-enabling in the 'if **name** == '**main**'' clause between the function calls. This seems very messy, I also used an implementation of the global keyword in my decorator which pylint doesn't like.
> 2. Context managers-
>    Because my original Lesson 05 assignment already used a context manager for mongo I couldn't really figure out what other useful things to add. I decided to incorporate a print statement for the **init**, **enter** and **exit** functions within the MongoDBConnection class. The **init** print statement seems the most useful as it prints out the port and host information when initialized.
> 3. Recursion-
>    This one was fun, I don't really have any questions on it, it was a good assignment and I'm always curious of more real world implementations of the type of recursion that python can do.
> 
> Thank you!
> 
> * Adam Strong

Hi Adam,

Good job on these three assignments.

On the decorator, it looks good. The way you're implementing it, you can set up a global level with the _debug_ parameter and then ON/OFF for decorated functions, which is useful (for example, you could prevent functions that have lots of log messages from crowding your log unless you need that information). If that is the case, your decorator should also restore logging to what it was once the decorated function has executed. 

About the context manager, the way you originally implemented it already included a context manager so there was not a lot to add there. Instead of print messages, I would add log messages, but what you're doing right now is already good for the assignment.

For recursion, yes, it's very useful. You have already implemented a real-world example, which is exploring a file directory. Other applications that use recursion include game engines (for example chess), determining the optimal route within a warehouse (which acts more or less like solving a labyrinth) or trip planning (the optimal route between points A and B considering available flights, schedule, number of stops, etc.).

I'll close the assignment for now, since you haven't submitted yet on the course website. Please open a new pull request once you're ready for grading.

Regards,

Luis

PR #840, opened 2020-08-01
Lesson 01 assignment package

bubblecfd:


ldconejo:
Very good first assignment, Pylint score of 10 and coverage higher than 90% for all files. Keep up the good work.

-Luis

PR #849, opened 2020-08-09
lesson4

Headieh:


students/headieh_godwin/lesson04/assignment/basic_operations.py
```diff
@@ -0,0 +1,122 @@
...
+            customer.save()
+            LOGGER.info("Datatbase updated successfully")
+    except IndexError:
+        LOGGER.info("customer_id does not exist")
+
+def list_active_customers():
+    """This function will return an integer with the number of customers
+    whose status is currently active."""
+    return len([customer.customer_id for customer in Customer.select().where(Customer.status)])
+    #active_count = Customer.select().where(Customer.status).count()
```
ldconejo, line 120: Always avoid leaving commented code (it's ok to do for a couple of hours while testing something). As a rule, someone else will uncomment it and cause a disaster :-)

PR #850, opened 2020-08-10
lesson 03 assignment files

bubblecfd:


ldconejo:
Hi Tao,

Looks good. Test coverage is below the 90% requirement. You are currently not testing any of the exceptions in your code, try to add tests for those (it's always a good idea for your testing to include fail flows). Let me know if you have any questions.

-Luis

bubblecfd:
Hi Luis,

I didn't see the *90%* requirement is mentioned anywhere in the assignment
description other than that tests are required. So I didn't check. But
anyway, I just checked, the coverage is 83% for basic_operations.py. So 7%
coverage short amounts to a 30% deduction in the grade? Thanks.

Tao Ye

ldconejo:
Hi Tao Ye,

The open edX tool, used for the assignments, does not have the option to give 80% or 90% of the grade. The next step after 100% is 70%. In any case, we believe good test coverage is very important in that it allows detection of, sometimes simple, errors which would otherwise make it into production, so it's a good habit to always make sure your tests cover at least 90% of your code (most projects in which I work require 100% coverage).

For the following assignments, please assume 90% test coverage requirement (if testing is required by the assignment) even if it is not explicitly stated.

Let me know if you have any questions on exception testing, it's an interesting topic, since it forces your code to crash.

Regards,

Luis

bubblecfd:
Hi Luis,

Wow, the grading system doesn't seem to be forgiving at all. One has to be
perfect to get 100%, right? Otherwise any small mistake would drop it all
the way to 70%. In my case, adding the coverage for the exception is little
work as I just did in a couple of minutes. The amount of the effort needed
certainly doesn't account for 30% of all work needed to get the
assignment done. Thanks.

ldconejo:
Hi Tao Ye,

Yes, that's a very good point. I've requested to have the option of entering the grade manually so that it better reflects what's missing in an assignment. On the positive side, I generally allow students to make additional submissions whenever necessary in order to get the full grade for an assignment. As you mentioned, in this case, a couple of minutes gets you to the full grade.

Thanks for the feedback,

Luis

bubblecfd:
Hi Luis,

I created a new pull request for the updated lesson 03 assignment
submission as pasted below

https://github.com/UWPCE-PythonCert-ClassRepos/SP_Python220B_2019/pull/854

Please let me know whether I need to also submit the zipped assignment
files.

Thank you.

PR #851, opened 2020-08-10
L4 V2

scotchwsplenda:
what is the difference between delete() with a where clause and delete_instance()?


ldconejo:
> what is the difference between delete() with a where clause and delete_instance()?

Check _Deleting records_ here for some code to illustrate the differences:

http://docs.peewee-orm.com/en/latest/peewee/querying.html

_delete_instance()_ is a method you can call for a record you have already identify. It's akin to telling a database record "delete yourself".

_delete()_ acts as a query, deleting any records that match.

PR #853, opened 2020-08-11
 Fix assignment 01, submit assignemnt 02 

lurveyt:
I liked assignment 2. I wasnt sure how for we should go in modifying the script but I went as far as I would have gone if I was writing this logging for someone else to use. The more verbose, the better it will be for them to find and fix their issues independently. I dont have a lot in my pdb log file for the assignment because the buffer was eaten up by all the debug prints.

Logging module seems to have several advantages over custom decorator loggers and I will probably integrate it into some of our code base at work. I also like pdb think it is a good option for quick debugging on a Raspberry Pi (not as much to setup as eclipse with remote debugging over the network.)


students/tim_lurvey/lesson02/assignment/code/charges_calc.py
```diff
@@ -0,0 +1,220 @@
...
+
+    parser.add_argument('-d', '--debug',
+                        help=debug_help_cmd,
+                        required=False,
+                        default='0')
+
+    return parser.parse_args()
+
+
+def set_logging_level():
```
ldconejo, line 39: For this one, try to consider the possibility of the user selecting an invalid logging level. Currently, the program crashes, which is fine (i.e., the fact that the user entered an invalid parameter will be noticed) but could be better handled using exceptions:

```
def set_logging_level():
    """
    Determine the logging level from the command line and set terminal and file logging.
    Default is no log file or debug terminal printing
    """
    log_level = {'0': {'term': logging.NOTSET,
                    'file': logging.NOTSET},
                '1': {'term': logging.ERROR,
                    'file': logging.ERROR},
                '2': {'term': logging.WARNING,
                    'file': logging.WARNING},
                '3': {'term': logging.DEBUG,
                    'file': logging.WARNING},
                }

    format_str = "%(asctime)s %(filename)s:%(lineno)-3d %(levelname)s %(message)s"

    try:
        log_terminal = logging.StreamHandler()
        log_terminal.setFormatter(fmt=logging.Formatter(format_str))
        log_terminal.setLevel(level=log_level.get(args.debug).get('term'))

        log_file = logging.FileHandler(filename=datetime.datetime.now().strftime("%Y-%m-%d") + ".log")
        log_file.setFormatter(fmt=logging.Formatter(format_str))
        log_file.setLevel(level=log_level.get(args.debug).get('file'))
    except AttributeError:
        print("Invalid logging level selected")
        exit()

    logger = logging.getLogger()
    logger.setLevel(logging.DEBUG)
    logger.addHandler(log_terminal)
    logger.addHandler(log_file)
```

PR #882, opened 2020-09-04
Lesson 07 Assignment

JavaRod:


students/rod_musser/lesson07/assignment/parallel.py
```diff
@@ -0,0 +1,230 @@
...
+    fields = [PRODUCT_FIELDS, CUSTOMER_FIELDS, RENTAL_FIELDS]
+    results = queue.Queue()
+
+    def worker(*args):
+        results.put(import_data(*args))
+
+    for i in range(3):
+        thread = threading.Thread(target=worker, args=('./data', data[i], tables[i], fields[i]))
+        thread.start()
+        threads.append(thread)
```
ldconejo, line 219: You are launching the threads, but you are not using any mechanism to ensure your threads have finished before you continue. Check the course materials on the _join()_ method.

ldconejo:
Hi Rod,

See me notes from a key element missing from your implementation. Also, please add your analysis on the results to _performance.txt_.

Thanks,

Luis

PR #885, opened 2020-09-06
Lesson 07 Assignment

JavaRod:
Added call to join() for threads created in parallel.py.  Added analysis to performance.txt


ldconejo:
Looks great now, thanks for the quick turnaround.

-Luis

PR #893, opened 2020-09-19
assignment 03

lurveyt:


students/tim_lurvey/lesson03/assignment/basic_operations.py
```diff
@@ -0,0 +1,81 @@
...
+                home_address=home_address,
+                phone_number=phone_number,
+                email_address=email_address,
+                status=status,
+                credit_limit=credit_limit,
+            )
+            new_customer.save()
+            logger.info(f'Customer record created: {customer_id}')
+
+    except Exception as e:
```
ldconejo, line 30: Please replace general exceptions with explicit exceptions. The problem with using general exceptions in released code is that they will captured everything, exceptions you have already seen and the ones you haven't, which may cause a new type of issue to go undetected or have unintended consequences because of improper handling.

students/tim_lurvey/lesson03/assignment/tests/test_unit.py
```diff
@@ -0,0 +1,3 @@
+#!/usr/env/bin python
```
ldconejo, line 1: Note that this file is empty. I assume this was in error, so please resubmit with the updated file.

ldconejo:
Hi Tim,

See my comments on your code regarding general exceptions. Also, make sure you resubmit with unit testing implemented. Let me know if you have any questions.

Regards,

Luis

